PR #2, opened 2021-06-04
Search feature data loading example

Windvis:


app/adapters/application.js
```diff
@@ -0,0 +1,4 @@
+import JSONAPIAdapter from '@ember-data/adapter/json-api';
+
+export default class ApplicationAdapter extends JSONAPIAdapter {
```
Windvis, line 3: Adding the adapter and serializer removes the deprecation message from the browser console.

app/controllers/people/index.js
```diff
@@ -1,28 +1,3 @@
 import Controller from '@ember/controller';
-import { tracked } from '@glimmer/tracking';
 
-export default class PeopleIndexController extends Controller {
-  @tracked shortedModel;
-
-  constructor() {
-    super(...arguments);
-    this.getAdminUnitInfo()
```
Windvis, line 9: The constructor can't be used for this since it's run only once, and not when the model hook resolves.

app/controllers/people/index.js
```diff
@@ -1,28 +1,3 @@
...
-  }
-
-  async getAdminUnitInfo() {
-    //let shortedModel = [];
-
-    const mandatories = await this.model.mandatories;
```
Windvis, line 15: Resolving relationships like this shouldn't be needed, and if you want to do it, I would do it in the model hook of the route instead.

app/models/mandate.js
```diff
@@ -2,6 +2,6 @@ import Model, {hasMany, belongsTo } from '@ember-data/model';
 
 export default class MandateModel extends Model {
     @belongsTo('board-position') roleBoard;
-    @belongsTo('governing-body', { inverse: true }) governingBody;
```
Windvis, line 5: `inverse: true` isn't valid. It should either be the name of the relationship in the linked model instead.

app/routes/people/index.js
```diff
@@ -3,6 +3,15 @@ import Route from '@ember/routing/route';
...
-export default class PeopleIndexRoute extends Route.extend(DataTableRouteMixin){ 
+export default class PeopleIndexRoute extends Route.extend(DataTableRouteMixin){
   modelName = 'person';
-}
\ No newline at end of file
+
+  mergeQueryOptions() {
+    return {
+      include: [
+        'mandatories.mandate.governing-body',
```
Windvis, line 12: Using includes reduces the amount of api calls when rendering the table. If you don't use includes, the table will make a call for each relationship per row so that adds up really fast. (and looks weird since the table updates once the data comes in).

Windvis: You can check the difference by commenting this code and checking the network inspector.

app/templates/people/index.hbs
```diff
@@ -37,14 +37,18 @@
...
                             @class="data-table__header-title" />
                         <AuDataTableThSortable @field="mandatories" @currentSorting={{this.sort}} @label="Organization"
                             @class="data-table__header-title" />
+                        <AuDataTableThSortable @field="mandatories.mandate.role-board.label" @currentSorting={{this.sort}} @label="Positie"
+                            @class="data-table__header-title" />
                     </c.header>
                     <c.body as |person|>
                         <td>{{person.givenName}}</td>
                         <td>{{person.familyName}}</td>
+                        <td>{{person.mandatories.firstObject.mandate.governingBody.name}}</td>
```
Windvis, line 50: This doesn't seem to contain data, so maybe I'm using the wrong relationships.

app/templates/people/index.hbs
```diff
@@ -37,14 +37,18 @@
...
+                        <td>{{person.mandatories.firstObject.mandate.roleBoard.label}}</td>
```
Windvis, line 51: I'm not sure if this is the right data to show, I'm not familiar with the data model, but it shows how you can simply use the relationships without having to do the logic that was added in the controller before.

PR #8, opened 2021-06-14
Administrative unit View and Edit

anitacaron:


app/models/worship-service.js
```diff
@@ -3,8 +3,17 @@ import AdministrativeUnitModel from './administrative-unit';
 
 export default class WorshipServiceModel extends AdministrativeUnitModel {
     @attr denomination;
-    @attr crossBorder;
+    @attr('boolean') crossBorder;
```
Windvis, line 6: I think it might be better to configure the backend to send a boolean instead? Or isn't that possible here?

tests/unit/controllers/administrative-units/administrative-unit-test.js
```diff
@@ -0,0 +1,12 @@
+import { module, test } from 'qunit';
```
Windvis, line 1: I would delete the test files. We don't write tests anyway :smile: and they sometimes cause false positives during CI.

Windvis: (I created a PR to delete the other ones already)

app/models/worship-service.js
anitacaron: Before only the backend was as boolean but I was not getting the correct value in the frontend. I discovered that in the database, it changed to 1 or 0 and not true or false, but it still shows 'undefined' in the ember add-on.

Windvis: Hmm, so for that to work a migration would be needed to convert it to proper booleans in the DB? I was hoping mu-cl-resources could do the conversion :smile:.

anitacaron: The TTL  file shows true/false, but when I search on the database, it shows 1/0. I don't know why it's changing.

Windvis: Maybe you can ask in the mu-semtech channel? Ideally this would be fixed in the backend. Using transforms is more of a last resort option imo.

anitacaron: It's not working in any case now. Maybe it's easier to create a migration SPARQL to change it, as you said. I've removed it.

Windvis: Oh it wasn't working with the transform either?

anitacaron: It was! This is causing much more problem than expected!

tests/unit/controllers/administrative-units/administrative-unit-test.js
Windvis: I think there are still 2 test files remaining for the routes?

anitacaron: Done!

PR #13, opened 2021-06-15
Reset the pagination when searching

Windvis:
I noticed the pagination wasn't being reset when applying a new filter (or resetting).

### DIFF
--- a/app/controllers/people/index.js
+++ b/app/controllers/people/index.js
@@ -31,6 +31,15 @@ export default class PeopleIndexController extends Controller {
   @action
   search(event) {
     event.preventDefault();
-    this.router.refresh();
+
+    if (this.page > 0) {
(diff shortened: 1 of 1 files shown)


app/controllers/people/index.js
```diff
@@ -17,6 +17,14 @@ export default class PeopleIndexController extends Controller {
 
   @action
   search() {
-    this.router.refresh();
+    if (this.page > 0) {
+      this.resetPagination(); // updating `page` will refresh the model
```
Windvis, line 21: This looks a bit weird, but updating the sort query param already reloads the model so refresh isn't needed here. Not sure how to make that more obvious without a comment.

anitacaron (approved):

Windvis:
I'll rebase, one second.

PR #16, opened 2021-06-16
Add a styled "no results" message to the people search page

Windvis:
Screenshot:
image


Windvis:
Woops, need to update the copy, this is the "people" version :smile:

PR #18, opened 2021-06-16
Add page titles to existing pages

Windvis:


app/templates/administrative-units.hbs
```diff
@@ -1,4 +1,5 @@
-{{breadcrumb "Besturen" route="administrative-units"}}
+{{page-title "Bestuurseenheden"}}
+{{breadcrumb "Bestuurseenheden" route="administrative-units"}}
```
Windvis, line 2: I didn't verify if this was ok, but I think it's weird to have a different breadcrumb title as the actual title on the page :smile:.

anitacaron: I think it's because it's following the design (?)

Windvis: Yea, I know. I meant I didn't verify with the design team :smile:.

PR #28, opened 2021-06-21
Prettier: Don't use single quotes for .hbs files

Windvis:
Using prettier for .hbs files should now keep the double quotes.


anitacaron:
Yay! Thanks for this PR!

anitacaron (approved):

PR #32, opened 2021-06-22
Add status colors

gcapurso:


app/components/administrative-unit-status.js
```diff
@@ -0,0 +1,13 @@
+import Component from '@glimmer/component';
+
+export default class AdministrativeUnitStatusComponent extends Component {
+  get getStatusSkin() {
+    if (this.args.status === 'Actief') {
```
Windvis, line 5: I would use the URI of the record instead of the label. I can imagine that the labels will change in the future while the URI _should_ always be the same, right?

An object which maps the URI to a skin value might be nice way to structure this:

```js
const ORGANIZATION_STATUS_SKINS = {
  'http://lblod.data.gift/concepts/abf4fee82019f88cf122f986830621ab': 'warning',
  // other types
}

export default class AdministrativeUnitStatusComponent extends Component {...}
```

And then you can simply return the correct skin by looking it up with the URI.

`return ORGANIZATION_STATUS_COLORS[someUriVariable];`

app/components/administrative-unit-status.js
```diff
@@ -0,0 +1,13 @@
+import Component from '@glimmer/component';
+
+export default class AdministrativeUnitStatusComponent extends Component {
+  get getStatusSkin() {
```
Windvis, line 4: nitpicky: `getStatusSkin` makes it sound like it's a function, but it's a getter. So I think simply using `statusSkin` might be clearer.

app/components/administrative-unit-status.js
```diff
@@ -0,0 +1,13 @@
+import Component from '@glimmer/component';
+
+export default class AdministrativeUnitStatusComponent extends Component {
```
Windvis, line 3: Optional: I know I pushed in this direction, but it seems the status is related to the organization, not "administrative unit". So renaming the component to `<OrganizationStatus>` might make sense?

app/components/administrative-unit-status.hbs
```diff
@@ -0,0 +1,3 @@
+<AuPill @skin={{this.getStatusSkin}}>
+  {{@status}}
```
Windvis, line 2: I think it might be better to pass in the full "organization-status-code" record and let the component handle the rendering instead of having to pass the specific label in.

Windvis (changes requested):

Windvis (approved):

PR #39, opened 2021-06-25
Change municipality/province from address and fix filter

anitacaron:


app/routes/administrative-units/index.js
```diff
@@ -29,12 +29,7 @@ export default class AdministrativeUnitsIndexRoute extends Route {
   @dropTask({ cancelOn: 'deactivate' })
   *loadAdministrativeUnitsTask(params) {
     let query = {
-      include: [
-        'classification',
-        'province',
-        'municipality',
-        'organization-status',
-      ].join(),
+      include: ['classification', 'organization-status', 'primary-site'].join(),
```
Windvis, line 32: I would include the address as well. `primary-site.address`. Otherwise Ember data is going to fetch that when used in the template and I hate that "data" popping in effect.

app/models/administrative-unit.js
```diff
@@ -3,8 +3,7 @@ import OrganizationModel from './organization';
 
 export default class AdministrativeUnitModel extends OrganizationModel {
   @belongsTo('administrative-unit-classification-code') classification;
-  @belongsTo('location') municipality;
-  @belongsTo('location') province;
+  @belongsTo('location') scope;
```
Windvis, line 6: What's this for?

Windvis (changes requested):

app/models/administrative-unit.js
anitacaron: It's the attribute for this 'Werkingsgebied' that is in the detail view.

Windvis (approved):

PR #41, opened 2021-06-28
Feature/view adminunit govbodies

gcapurso:


app/controllers/administrative-units/administrative-unit/governing-bodies/index.js
```diff
@@ -0,0 +1,18 @@
+import Controller from '@ember/controller';
+
+export default class AdministrativeUnitsAdministrativeUnitGoverningBodiesIndexController extends Controller {
+  get governingBodies() {
```
Windvis, line 4: It seems this isn't used anymore. The template uses the model directly. So I would just remove the controller if it's not used for anything else.

app/routes/administrative-units/administrative-unit/governing-bodies/index.js
```diff
@@ -0,0 +1,26 @@
...
+      administrativeUnitId,
+      {
+        reload: true,
+        include: 'governing-bodies.has-time-specializations',
+      }
+    );
+
+    return {
+      administrativeUnit: administrativeUnit,
+      governingBody: await administrativeUnit.governingBodies.firstObject,
```
Windvis, line 23: I would add a comment about why only the first one is used. If I understood it correctly worship services related administrative units only have 1 governing body?

app/routes/administrative-units/administrative-unit/governing-bodies/index.js
```diff
@@ -0,0 +1,26 @@
...
+    let { id: administrativeUnitId } = this.paramsFor(
+      'administrative-units.administrative-unit'
+    );
+
+    let administrativeUnit = await this.store.findRecord(
+      'administrative-unit',
+      administrativeUnitId,
+      {
+        reload: true,
+        include: 'governing-bodies.has-time-specializations',
```
Windvis, line 17: I would also add the `governing-bodies.classification` relationship since it's fetched async now.

Windvis (changes requested):

app/routes/administrative-units/administrative-unit/governing-bodies/index.js
gcapurso: Yes exactly, and then tthere are many gpverning bodies nested

Windvis (approved):

PR #50, opened 2021-07-06
Add the `<MunicipalitySelect>` and `<ProvinceSelect>` components in the filter.

anitacaron:


app/templates/administrative-units/index.hbs
```diff
@@ -10,14 +10,22 @@
           </p>
           <p>
             <AuLabel for="filter-municipality">Gemeente</AuLabel>
-            <AuInput
-              id="filter-municipality"
-              @width="block"
-              @value={{this.municipality}}
+            <MunicipalitySelect
+              @location="municipality"
```
Windvis, line 14: I think this isn't used anymore?

app/templates/administrative-units/index.hbs
```diff
@@ -10,14 +10,22 @@
...
+              @selected={{this.municipality}}
+              @set={{this.setMunicipality}}
             />
           </p>
           <p>
-            <AuLabel for="classifications">Soort eredienst</AuLabel>
+            <AuLabel for="filter-province">Provincie</AuLabel>
+            <ProvinceSelect
+              @location="province"
```
Windvis, line 22: Same here.

app/templates/administrative-units/index.hbs
```diff
@@ -10,14 +10,22 @@
...
           <p>
             <AuLabel for="filter-municipality">Gemeente</AuLabel>
-            <AuInput
-              id="filter-municipality"
-              @width="block"
-              @value={{this.municipality}}
+            <MunicipalitySelect
+              @location="municipality"
+              @selected={{this.municipality}}
+              @set={{this.setMunicipality}}
```
Windvis, line 16: Nitpicky, but I prefer if actions have a more descriptive name. I usually go for `onChange` if I want to keep it generic.

tests/integration/components/municipality-select-test.js
```diff
@@ -0,0 +1,26 @@
+import { module, test } from 'qunit';
```
Windvis, line 1: The tests can be removed.

app/components/municipality-select.js
```diff
@@ -0,0 +1,34 @@
...
+export default class MunicipalitySelectComponent extends Component {
+  @service fastboot;
+  @service store;
+  municipalities;
+
+  constructor(...args) {
+    super(...args);
+
+    if (!this.fastboot.isFastBoot) {
+      this.municipalities = this.loadMunicipalitiesTask.perform();
```
Windvis, line 14: This stores the TaskInstance instead of the actual data. TaskInstances act like promises, they can resolve to a value if that's returned from the task.

app/components/municipality-select.js
```diff
@@ -0,0 +1,34 @@
...
+      },
+      page: {
+        size: 400,
+      },
+      sort: 'label',
+    };
+
+    const municipalities = yield this.store.query('location', query);
+
+    return municipalities.mapBy('label');
```
Windvis, line 32: Returning the data here, makes it available when the "promise" resolves. Which PowerSelect knows how to handle.

app/components/municipality-select.hbs
```diff
@@ -0,0 +1,14 @@
+<PowerSelect
+  @allowClear={{true}}
+  @searchEnabled={{true}}
+  @loadingMessage="Aan het laden..."
+  @noMatchesMessage="Geen resultaten"
+  @searchMessage="Typ om te zoeken"
+  @options={{this.municipalities}}
```
Windvis, line 7: This is the TaskInstance we set earlier. If the promise is running, PowerSelect will show the loadingMessage when opened. Once the promise resolves the options will be available.

app/components/province-select.js
```diff
@@ -0,0 +1,32 @@
...
+import { restartableTask } from 'ember-concurrency';
+import { tracked } from '@glimmer/tracking';
+
+export default class ProvinceSelectComponent extends Component {
+  @service store;
+  @tracked provinces;
+
+  constructor(...args) {
+    super(...args);
+    this.loadProvinces.perform();
```
Windvis, line 12: I would mimic the updated setup in the MunicipalitySelect component.

Windvis (changes requested):
Awesome!

Windvis (approved):
Good stuffs!

PR #53, opened 2021-07-07
Add sites details page

gcapurso:


app/templates/administrative-units/administrative-unit/sites/site/index.hbs
```diff
@@ -0,0 +1,100 @@
+<div class="au-o-box">
+  <AuToolbar>
+    <AuToolbarGroup>
+      <AuHeading>
+        Vestiging: {{@model.sites.address.fullAddress}}
+      </AuHeading>
+    </AuToolbarGroup>
+    <AuToolbarGroup>`
```
Windvis, line 8: There is ` character there :smile: I thought it was a little insect on my screen at first :grimacing:

app/templates/administrative-units/administrative-unit/sites/site/index.hbs
```diff
@@ -0,0 +1,100 @@
...
+              </div>
+            </div>
+          </div>
+          <div class="au-o-grid__item au-u-1-1 au-u-1-2@medium">
+            <div class="au-o-grid">
+              <div class="au-o-grid__item au-u-1-4">
+                <p><strong>Telefoonnummer</strong></p>
+              </div>
+              <div class="au-o-grid__item au-u-3-4">
+                <p>{{@model.sites.contacts.firstObject.telephone}}</p>
```
Windvis, line 82: I think most of these values can be empty, so I would wrap them in a check to not show the label if there is no value.

app/routes/administrative-units/administrative-unit/sites/site/index.js
```diff
@@ -0,0 +1,33 @@
...
+          'primary-site.contacts',
+          'sites.address',
+          'sites.contacts',
+        ].join(),
+      }
+    );
+
+    return {
+      administrativeUnit: administrativeUnit,
+      primarySite: await administrativeUnit.primarySite,
```
Windvis, line 28: I think this can be replaced with a simple boolean `isPrimarySite` and do the comparison that you do in the template here instead. The primarySite record isn't needed in the template for anything else I think.

Windvis (changes requested):

app/routes/administrative-units/administrative-unit/sites/site/index.js
```diff
@@ -0,0 +1,33 @@
...
+
+export default class AdministrativeUnitsAdministrativeUnitSitesSiteIndexRoute extends Route {
+  @service store;
+
+  async model() {
+    let { id: administrativeUnitId } = this.paramsFor(
+      'administrative-units.administrative-unit'
+    );
+
+    let administrativeUnit = await this.store.findRecord(
```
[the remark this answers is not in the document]
gcapurso: Yes but we need to show the administative page name too, that's why I include that too. It works even in that case?

app/templates/administrative-units/administrative-unit/sites/site/index.hbs
gcapurso: ahahahahah

app/routes/administrative-units/administrative-unit/sites/site/index.js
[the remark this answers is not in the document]
Windvis: Yea, but we already fetch the administrative unit in one of the parent routes I think, so you can retrieve that with the [modelFor](https://api.emberjs.com/ember/3.27/classes/Route/methods/modelFor?anchor=modelFor) method.

This should work I think:
```js
let administrativeUnit = this.modelFor('administrative-units.administrative-unit');
```

app/routes/administrative-units/administrative-unit/sites/site/index.js
gcapurso: I solved it in a different way, by accessing to @model.administrativeUnit.primarySite in the template (we should include administrativeUnit in any cas fot the nale as discussed above).

app/routes/administrative-units/administrative-unit/sites/site.js
```diff
@@ -0,0 +1,25 @@
...
+import { inject as service } from '@ember/service';
+export default class AdministrativeUnitsAdministrativeUnitSitesSiteRoute extends Route {
+  @service store;
+
+  async model() {
+    let administrativeUnit = this.modelFor(
+      'administrative-units.administrative-unit'
+    );
+
+    let { siteId: siteId } = this.paramsFor(
```
Windvis, line 11: paramsFor shouldn't be needed anymore, you can now directly access the params inside the model hook. The first argument to the is an object with all the params.

->
`async model({ siteId }) { //... }`

app/templates/administrative-units/administrative-unit/sites.hbs
```diff
@@ -1,4 +1,4 @@
-{{page-title "Vestigingen"}}
+{{page-title "Vestiging"}}
```
Windvis, line 1: Should be "Vestigingen" I think, as it was before?

app/templates/administrative-units/administrative-unit/sites/index.hbs
```diff
@@ -25,9 +25,15 @@
       <th></th>
     </c.header>
     <c.body as |site|>
-      <td><AuLink @route="administrative-units.administrative-unit.sites">{{site.address.fullAddress}}</AuLink></td>
+      <td>
+        <AuLink @model={{site.id}} @route="administrative-units.administrative-unit.sites.site.index">
```
Windvis, line 29: Nitpicky: It's usually better to not link to an index route directly. That's an implementation detail of that route. The parent route doesn't need to know that `administrative-units.administrative-unit.sites.site` uses an index route.

I would change this to `@route="administrative-units.administrative-unit.sites.site"`

app/templates/administrative-units/administrative-unit/sites/site.hbs
```diff
@@ -0,0 +1,11 @@
+{{page-title @model.site.address.fullAddress}}
+
+{{breadcrumb
+  this.model.site.address.fullAddress
+  route="administrative-units.administrative-unit.sites"
+  models=(array this.model.site.id this.model.administrativeUnit.id)
```
Windvis, line 6: I think the order needs to be reversed here: admin unit id first, site id second. I think ember applies them from left to right.

Windvis (changes requested):
Last batch of small things :grinning:

app/routes/administrative-units/administrative-unit/sites/site.js
gcapurso: Can I have more details about the reason why is not anymore necessary?

Windvis: Sure. The router.js file shows that the `:siteId` dynamic segment was added to the `administrative-units.administrative-unit.sites.site` route. This means that that route will receive the value for that param as the first argument to the model hook (which is what I suggested here).

Before, `paramsFor` was needed since the fetching was done from the `administrative-units.administrative-unit.sites.site.index` route. That's not the route that defines the param so we don't receive it directly in the model hook.

gcapurso: Clear! Thank you!

app/templates/administrative-units/administrative-unit/sites/site.hbs
```diff
@@ -0,0 +1,11 @@
+{{page-title @model.site.address.fullAddress}}
+
+{{breadcrumb
+  this.model.site.address.fullAddress
+  route="administrative-units.administrative-unit.sites"
+  models=(this.model.administrativeUnit.id array this.model.site.id)
```
Windvis, line 6: Wait, doesn't this throw an error at the moment? :smile: array is a helper so it should be the first thing there.

`(array this.model.administrativeUnit.id this.model.site.id)`

Windvis (approved):
Nice! :tada:

PR #54, opened 2021-07-07
Fix breadcrumbs design

gcapurso:


Windvis:
This makes no sense to me, please discuss things like that before applying the changes.

Windvis:
I just noticed how harsh my wording in the previous message was. I blame it on not being a morning person and having a bad nights sleep, but that doesn't justify it of course :slightly_smiling_face:. I'll work on my communication. Sorry about that.

PR #63, opened 2021-07-13
Administrative unit core data updates

Windvis:
- use the new `<DataCard>` component
- show the correct classification

### DIFF
--- a/app/templates/administrative-units/administrative-unit/core-data/index.hbs
+++ b/app/templates/administrative-units/administrative-unit/core-data/index.hbs
@@ -25,240 +25,133 @@
       </AuHeading>
     </div>
 
-    <div>
-      <AuToolbar class="au-u-margin-bottom-small">
-        <AuToolbarGroup>
-          <AuHeading @level="4" @skin="5">Bestuur van de eredienst</AuHeading>
-        </AuToolbarGroup>
-      </AuToolbar>
-      <AuCard @standOut={{true}} as |c|>
-        <c.content>
-          <div class="au-flow au-o-flow--large">
-            <div class="au-o-grid">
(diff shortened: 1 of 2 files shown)


app/templates/administrative-units/administrative-unit/core-data/index.hbs
```diff
@@ -25,240 +25,131 @@
...
-                  </div>
-                </div>
-              {{/each}}
-            </div>
-          </div>
-        </c.content>
-      </AuCard>
-    </div>
+    <DataCard>
+      <:title>{{@model.administrativeUnit.classification.label}}</:title>
```
Windvis, line 29: This was hardcoded to "Bestuur van de eredienst" before.

app/templates/administrative-units/administrative-unit/core-data/index.hbs
```diff
@@ -25,240 +25,131 @@
...
+            <:content>{{@model.administrativeUnit.recognizedWorshipType.label}}</:content>
+          </Item>
+          {{#if @model.administrativeUnit.denomination}}
+            <Item>
+              <:label>Strekking</:label>
+              <:content>{{@model.administrativeUnit.denomination}}</:content>
+            </Item>
+          {{/if}}
+        </Grid>
+
```
anitacaron, line 51: Cross border only exists for worship service, it's better if we include the conditional here?

anitacaron (changes requested):

app/templates/administrative-units/administrative-unit/core-data/index.hbs
Windvis: Ah, yea I can add that as well :+1:

anitacaron (approved):

PR #66, opened 2021-07-13
Site edit page

anitacaron:


anitacaron:
Still missing the site type 😬

app/routes/administrative-units/administrative-unit/sites/site/edit.js
```diff
@@ -0,0 +1,21 @@
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+
+export default class AdministrativeUnitsAdministrativeUnitSitesSiteEditRoute extends Route {
+  @service store;
+
+  async model() {
+    let adminUnitSite = this.modelFor(
```
Windvis, line 8: This was a bit confusing. I assumed it was the site record while it actually is the model of the parent. Returning the object directly would be clearer I think. `this.model.adminUnitSite.site` -> `this.model.site`

app/components/site-type-select.hbs
```diff
@@ -0,0 +1,12 @@
+<PowerSelect
+  @allowClear={{true}}
+  @loadingMessage="Aan het laden..."
+  @noMatchesMessage="Geen resultaten"
+  @searchMessage="Typ om te zoeken"
```
Windvis, line 5: The search arguments don't really do anything at the moment. `@searchEnabled={{true}}` activates it (but the `searchField` needs to be configured as well).

app/components/site-type-select.js
```diff
@@ -0,0 +1,22 @@
...
+    super(...args);
+
+    if (!this.fastboot.isFastBoot) {
+      this.siteTypes = this.loadSiteTypesTask.perform();
+    }
+  }
+
+  @task
+  *loadSiteTypesTask() {
+    return yield this.store.findAll('site-type');
```
Windvis, line 20: I would pass the [`reload: true`](https://api.emberjs.com/ember-data/3.27/classes/Store/methods/query?anchor=findAll) option to findAll. By default store.findAll returns records from the store if some are loaded already. This sounds nice on paper, but in practise that results in weird behavior:

Let's say the user loads the app on a specific site details page. That details page shows the site type relationship so the store will have 1 record of the `site-type` model. If the user then goes to the edit page, this call will instantly resolve to the 1 record of the store (and it will do a fetch in the background to find all of them).

So in that scenario the select will have only 1 option until the background request completes, which looks like a bug to end users.

Passing reload forces the fetch to be resolved before returning any data.

app/controllers/administrative-units/administrative-unit/sites/site/edit.js
```diff
@@ -0,0 +1,76 @@
+import Controller from '@ember/controller';
+import { dropTask } from 'ember-concurrency';
+import { inject as service } from '@ember/service';
+import { tracked } from '@glimmer/tracking';
+
+export default class AdministrativeUnitsAdministrativeUnitSitesSiteEditController extends Controller {
+  @service router;
+  @tracked isPrimarySite =
+    this.site.get('id') === this.administrativeUnit.get('primarySite.id');
```
Windvis, line 9: Wait, does this work? :smile:  The controller shouldn't have access to these values when it's created and it won't rerun when new data arrives through the model hook.

Changing this into a getter should be work for sure.

app/controllers/administrative-units/administrative-unit/sites/site/edit.js
```diff
@@ -0,0 +1,76 @@
...
+    return this.model.adminUnitSite.administrativeUnit;
+  }
+
+  @dropTask
+  *save(event) {
+    event.preventDefault();
+
+    let address = yield this.site.address;
+
+    if (address.hasDirtyAttributes) {
```
Windvis, line 25: Can you add a TODO comment here? I think we should make this reusable since we'll need it on multiple edit pages. I still need to think of a good way to tackle that though :smile:

app/templates/administrative-units/administrative-unit/sites/site/edit.hbs
```diff
@@ -0,0 +1,169 @@
...
+      </AuHeading>
+      <AuHeading @level="2" @skin="3" class="au-u-margin-bottom-tiny">
+        {{@model.adminUnitSite.site.address.fullAddress}},
+        {{@model.adminUnitSite.administrativeUnit.name}}
+      </AuHeading>
+    </div>
+
+    <form
+      id="edit-vestiging-form"
+      aria-label="edit-vestiging-form"
```
Windvis, line 15: Is that something that's forced by a linting rule or something? First time I see it :smile:

app/controllers/administrative-units/administrative-unit/sites/site/edit.js
Windvis: Ah wait, I see now, it's also used to store the checkbox value. Hmmm. I think it makes sense to have easy access to the value before it was edited.

Maybe adding an `isPrimaryCurrently` or something better named might be nice?

app/controllers/administrative-units/administrative-unit/sites/site/edit.js
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    let contacts = yield this.site.contacts;
+    if (contacts.firstObject.hasDirtyAttributes) {
+      yield contacts.firstObject.save();
+    }
+
+    yield this.site.save();
+
+    if (!this.isPrimarySite) {
```
Windvis, line 52: I think this primary site logic isn't 100% solid yet. If the user checks the edit page of a non-primary site, and saves the site, it will remove the actual (other) primary site relationship, right? So after that the admin unit won't have a primary site anymore? I might be missing something though.

Windvis: I think this should check if the site was the primary site _before_ the user entered the page and if it was, only then remove the primary site relationship?

app/controllers/administrative-units/administrative-unit/sites/site/edit.js
```diff
@@ -0,0 +1,76 @@
...
+      this.administrativeUnit.sites.pushObject(this.site);
```
Windvis, line 53: I still think it makes more sense to store _all_ sites in the sites relationship, so also the one that is marked the primary-site. That way we only need to add or remove the primary site relationship without touching the `sites` one. It just seems weird to not get the primary site when accessing the sites relationship.

Windvis (changes requested):

app/routes/administrative-units/administrative-unit/sites/site/edit.js
anitacaron: But I also need the administrative unit model.

app/templates/administrative-units/administrative-unit/sites/site/edit.hbs
anitacaron: I use `Ctrl+Shift+I` to format

app/components/site-type-select.hbs
anitacaron: No need the search now, I'll delete the message.

app/components/site-type-select.js
anitacaron: Thanks for the explanation! 😄

app/templates/administrative-units/administrative-unit/sites/site/edit.hbs
Windvis: I meant the aria-label :smile:

app/routes/administrative-units/administrative-unit/sites/site/edit.js
Windvis: Yea but that's not an issue right? I just mean that instead of doing:

```js
return { adminUnitSite }
```

You could do:
```js
return adminUnitSite
```
To remove the extra object wrapper.

app/templates/administrative-units/administrative-unit/sites/site/edit.hbs
anitacaron: It's like this in the core edit 😬

Windvis: ahh, we probably had multiple form elements on that page at one point. Anyhow, doesn't really matter.

app/routes/administrative-units/administrative-unit/sites/site/edit.js
anitacaron: I'm a bit slow today. 😬

app/controllers/administrative-units/administrative-unit/sites/site/edit.js
anitacaron: I need to change the data, then.

Windvis: Yea, maybe we should discuss that with other people first. I didn't define the data model so my expectations might not match the reallity :smile:.

app/controllers/administrative-units/administrative-unit/sites/site/edit.js
anitacaron: Yes, I forgot about that use case.

Windvis: Just a summary to see if I understood it correctly:

Keeping the current data model in mind (primary-site isn't part of sites) the following scenario's are possible:

1. The user edits the site that is also the primary site at the moment:
    1. It's still marked as primary site on save -> nothing needs to move around
    2. It's not marked as primary anymore
        - primary site relationship needs to be cleared
        - the site record needs to be added to the sites relationship
        - (This scenario is problematic since we use the primary site in multiple places :smile:.)
2. The user edits a site that isn't the primary site at the moment
    1. It's now marked as the primary site on save 
        - the record that's currently linked as primary site needs to be added to the sites relationship
        - the record that's being edited needs to be set as the primary site record and removed from the sites relationship
     2. It's not marked as primary on save -> nothings needs to move around

anitacaron: If we show the info we have, I think it won't be a problem. If the user doesn't select a primary site, it's not a bug.

anitacaron: To confirm, I need to change the first if only to change it if before it was a primary site.

Windvis (approved):

PR #67, opened 2021-07-15
Add contact information and move edit logic

gcapurso:


app/templates/people/person/personal-information/edit.hbs
```diff
@@ -0,0 +1,101 @@
...
+              <div class="au-o-grid">
+                <div class="au-o-grid__item au-u-1-1 au-u-1-2@medium">
+                  <div class="au-o-grid">
+                    <div class="au-o-grid__item au-u-1-4">
+                      <AuLabel for="person-given-name">
+                        Voornaam
+                      </AuLabel>
+                    </div>
+                    <div class="au-o-grid__item au-u-3-4">
+                      <AuInput @value={{@model.givenName}} @width="block" id="person-given-name" />
```
Windvis, line 38: using `@model` as a record is confusing. I would either return an object with a person key from the model hook (so this becomes `@model.person.givenName` or alias the model in the controller so it becomes `this.person.givenName`. That will make things way more readable imo.

Windvis (changes requested):

app/templates/people/person/personal-information/index.hbs
```diff
@@ -1,224 +1,93 @@
...
+            <:content>{{@model.person.mandatories.firstObject.contacts.firstObject.telephone}}</:content>
+          </Item>
+        </Grid>
+        <Grid as |Item|>
+          <Item>
+            <:label>Adres</:label>
+            <:content>{{@model.person.mandatories.firstObject.contacts.firstObject.contactAddress.fullAddress}}</:content>
+          </Item>
+          <Item>
+            <:label>GSM nummer</:label>
```
Windvis, line 86: Since we don't have data for this yet I would hide it for now maybe?

app/templates/people/person/personal-information/index.hbs
```diff
@@ -0,0 +1,93 @@
...
+    <DataCard>
+      <:title>Persoonlijke gegevens</:title>
+      <:card as |Grid|>
+        <Grid as |Item|>
+          <Item>
+            <:label>Voornaam</:label>
+            <:content>{{@model.person.givenName}}</:content>
+          </Item>
+        </Grid>
+        <Grid as |Item|>
```
Windvis, line 35: These can all be added under the same `<Grid>` component. A new `<Grid>` is only needed if you want to force things on a new line.

app/templates/people/person/personal-information/index.hbs
```diff
@@ -0,0 +1,93 @@
...
+            <:content>{{@model.person.mandatories.firstObject.contacts.firstObject.email}}</:content>
+          </Item>
+          <Item>
+            <:label>Telefoonnummer</:label>
+            <:content>{{@model.person.mandatories.firstObject.contacts.firstObject.telephone}}</:content>
+          </Item>
+        </Grid>
+        <Grid as |Item|>
+          <Item>
+            <:label>Adres</:label>
```
Windvis, line 82: I think the address can be added to the same `<Grid>` as the email and phone number, right?

app/templates/people/person/personal-information/index.hbs
```diff
@@ -0,0 +1,93 @@
...
+              <AuLink
+                @model={{@model.person.mandatories.firstObject.mandate.governingBody.isTimeSpecializationOf.administrativeUnit.id}}
+                @route="administrative-units.administrative-unit">
+                {{@model.person.mandatories.firstObject.mandate.governingBody.isTimeSpecializationOf.administrativeUnit.name}}
+              </AuLink>
+            </:content>
+          </Item>
+        </Grid>
+        <Grid as |Item|>
+          <Item>
```
Windvis, line 71: A lot of these contact thingies seem empty (for the people I checked at least), so we shouldn't show the label either in that case. I would wrap all of them in if checks.

app/templates/people/person/personal-information/edit.hbs
```diff
@@ -0,0 +1,101 @@
...
+
+      <div class="au-o-flow au-o-flow--large">
+
+        <div>
+          <AuToolbar class="au-u-margin-bottom-small">
+            <AuToolbarGroup>
+              <AuHeading @level="3" @skin="5">Persoonlijke gegevens</AuHeading>
+            </AuToolbarGroup>
+          </AuToolbar>
+          <AuCard as |c|>
```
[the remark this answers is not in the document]
Windvis: This was resolved, but the edit page wasn't updated yet I think :smile:?

Windvis (changes requested):

app/templates/people/person/personal-information/index.hbs
gcapurso: Ok :)

app/templates/people/person/personal-information/index.hbs
gcapurso: Ok but in that case to keep the layout I have to use different grids.

app/templates/people/person/personal-information/index.hbs
anitacaron: Yes, we have this!

anitacaron: Sorry, I thought it was the address!

app/templates/people/person/personal-information/index.hbs
gcapurso: We should use a different grid if we make contact fields optional throught the if. Otherwise it will change position when one attribute is missing.

Windvis: That doesn't really matter right? The only problem is that if all items in a grid are empty / hidden it looks weird, so that probably requires a large check as well. But I guess that won't happen in practice.

gcapurso: Maybe we can check on @model.person.mandatories.firstObject?

app/templates/people/person/personal-information/index.hbs
```diff
@@ -0,0 +1,110 @@
...
+            <:label>Achternaam</:label>
+            <:content>{{@model.person.familyName}}</:content>
+          </Item>
+        </Grid>
+        {{#if @model.person.firstNameUsed}}
+          <Item>
```
Windvis, line 43: This needs to be nested under a grid component. I don't think it's valid at the moment.

app/templates/people/person/personal-information/index.hbs
```diff
@@ -0,0 +1,110 @@
...
+          }}
+            <Item>
+              <:label>Telefoonnummer</:label>
+              <:content>
+                {{@model.person.mandatories.firstObject.contacts.firstObject.telephone}}
+              </:content>
+            </Item>
+          {{/if}}
+        </Grid>
+        <Grid as |Item|>
```
Windvis, line 91: I think these can be part of the email and phone grid, right?

app/templates/people/person/personal-information/index.hbs
anitacaron: This is the design, no? Because the second column will be the date of birth and RRN, no?

Windvis: I assumed the items would just flow through to the next line if needed. Forcing items on a new line because it's there in the design isn't the best reason I think. The other data isn't there yet so in my head the last name just moves to the first row.

anitacaron: Yes, but they're expected to be on the left side; having things moving depending on other attributes is not consistent on all pages.

Windvis: Well it's only my opinion, showing them in new row while the right column is empty looks weird to me. Anyway, leave it as is then.

PR #73, opened 2021-07-26
Refactor/card columns

Windvis:
This adds a Columns option to the cards components:

```hbs
        <EditCard>
          <:title>Bestuurseenheid</:title>
          <:card as |Card|>
            <Card.Columns>
             <:left as |Item|>...</:left>
             <:right as |Item|>...</:right>
           </Card.Columns>
        </EditCard>
```

This is a breaking change since the previous version did not yield an object. I fixed this for all the existing code, but branches will need to be updated:

before:
```hbs
        <EditCard>
          <:title>Bestuurseenheid</:title>
          <:card as |Grid|>
            <Grid as |Item|>
             ...
           </Grid>
        </EditCard>
```

after:
```hbs
        <EditCard>
          <:title>Bestuurseenheid</:title>
          <:card as |Card|>
            <Card.Grid as |Item|>
            ...
           </Card.Grid>
        </EditCard>
```

### DIFF
--- a/app/components/card-columns.hbs
+++ b/app/components/card-columns.hbs
@@ -1,0 +1,16 @@
+<div class="card-columns" ...attributes>
+  <div class="card-columns__column">
+    {{#if @itemComponent}}
+      {{yield (component @itemComponent) to="left"}}
+    {{else}}
+      {{yield to="left"}}
+    {{/if}}
+  </div>
+  <div class="card-columns__column">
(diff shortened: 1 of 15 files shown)


Windvis:
We can remove the Grid component once we converted everything to the Columns version. But this way there is no pressure to change everything at once.

anitacaron (approved):

PR #81, opened 2021-07-28
Refactor/use page header component

Windvis:
This uses the PageHeader on all pages that had the custom title structures.


Windvis:
I'll rebase and update the mandatory edit page as well.

anitacaron (approved):

PR #82, opened 2021-07-29
Landingspagina before login

anitacaron:


app/templates/login.hbs
```diff
@@ -0,0 +1,141 @@
...
+      <AuContentHeader
+        @titlePartOne="Vlaanderen"
+        @titlePartTwo="is lokaal bestuur"
+      >
+        <img
+          sizes="50vw"
+          src={{root-url
+            "/assets/images/loket-header-1600-85a5251e74ef6ac8491b2f228d2c90da.jpg"
+          }}
+          srcset={{@srcSet}}
```
anitacaron, line 13: It's not showing the image for me.

app/templates/login.hbs
```diff
@@ -0,0 +1,141 @@
...
+              <AuButton @icon="login" @iconAlignment="left">
+                Aanmelden
+              </AuButton>
+            </div>
+          </div>
+        </div>
+      </section>
+
+      <section class="au-o-region-large au-u-background-gray-200">
+        <div class="au-o-layout">
```
anitacaron, line 49: Weird that the background, for me, is white.

app/templates/login.hbs
Windvis: It needs to be copied into the public assets folder first.

app/templates/login.hbs
Windvis: I think this is a new util class. Maybe we need to import the correct sass file first.

app/templates/login.hbs
```diff
@@ -0,0 +1,141 @@
+<AuApp>
```
Windvis, line 1: I have to double check what this outputs, but I think this isn't  needed since we already have it in the application.hbs template. Same with the main container.

app/templates/login.hbs
Windvis: and the `@srcSet` will be undefined here. It needs to be hardcoded path similar to the source.

app/templates/login.hbs
anitacaron: Can I find which file I need to import?

Windvis: https://github.com/appuniversum/ember-appuniversum/commit/599217331c1f78b2ae33d1c6c1dff3e75e9fddc7#diff-df64b11c1686e6f94fe51273b886b615366586da01de003b9081373bacd726f7R98
Although the others are probably needed as well.

anitacaron: Thanks!

app/templates/login.hbs
```diff
@@ -0,0 +1,141 @@
...
+          src={{root-url
+            "/assets/images/loket-header-1600-85a5251e74ef6ac8491b2f228d2c90da.jpg"
+          }}
+          srcset={{@srcSet}}
+          alt="Foto van een laptop met daarop het vlaanderen logo."
+        />
+      </AuContentHeader>
+
+      <section class="au-o-region-large">
+        <div class="au-o-layout">
```
Windvis: These can be combined into 1 element and  replaced by the `au-o-box au-o-box--large` classes.

app/templates/login.hbs
```diff
@@ -0,0 +1,141 @@
...
+            >
+              <AuButton @icon="login" @iconAlignment="left">
+                Aanmelden
+              </AuButton>
+            </div>
+          </div>
+        </div>
+      </section>
+
+      <section class="au-o-region-large au-u-background-gray-200">
```
Windvis: Same here (`au-o-box au-o-box--large`)

app/templates/login.hbs
```diff
@@ -0,0 +1,141 @@
...
+        <div class="au-o-layout">
+          <AuContent @skin="large" class="au-u-margin-bottom">
```
Windvis: The AuContent wrapper isn't needed.

app/templates/login.hbs
Windvis: Yea, they aren't needed here.

Windvis (changes requested):

app/templates/login.hbs
```diff
@@ -0,0 +1,141 @@
+<AuApp>
+  <AuMainContainer as |m|>
+    <m.content @scroll={{true}}>
+      <AuContentHeader
+        @titlePartOne="Vlaanderen"
+        @titlePartTwo="is lokaal bestuur"
```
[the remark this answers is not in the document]
anitacaron: 😬

app/templates/login.hbs
```diff
@@ -0,0 +1,141 @@
...
+      >
+        <img
+          sizes="50vw"
+          src={{root-url
+            "/assets/images/loket-header-1600-85a5251e74ef6ac8491b2f228d2c90da.jpg"
+          }}
+          srcset={{@srcSet}}
+          alt="Foto van een laptop met daarop het vlaanderen logo."
```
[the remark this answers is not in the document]
anitacaron: It's good for the accessibility, no?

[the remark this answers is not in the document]
Windvis: A screenreader would read that, yes. But I don't think visually impaired people wants / needs to know that to use the website. Just adding alt texts everywhere is the opposite of accessibility imo :smile:.

PR #85, opened 2021-07-30
Add explicit inverse relationship names

Windvis:
I noticed a missing inverse related error message when clicking through the administrative units search pages so I decided to add it to all the relationships. Most of the changes are just formatting though.

More info: https://github.com/emberjs/rfcs/pull/739

Error message:
```
Uncaught Error: Assertion Failed: You defined the 'involvedBoards' relationship on model:worship-service, but you defined the inverse relationships of type model:local-involvement multiple times. Look at https://guides.emberjs.com/current/models/relationships/#toc_explicit-inverses for how to explicitly specify inverses
```

### DIFF
--- a/app/models/post.js
+++ b/app/models/post.js
@@ -1,7 +1,18 @@
 import Model, { belongsTo, hasMany } from '@ember-data/model';
 
 export default class PostModel extends Model {
-  @belongsTo('role') generalRole;
-  @belongsTo('organization') organization;
-  @hasMany('agent-in-position', { inverse: 'position' }) agentsInPosition;
+  @belongsTo('role', {
+    inverse: null,
+  })
+  generalRole;
+
+  @belongsTo('organization', {
+    inverse: null,
+  })
+  organization;
+
+  @hasMany('agent-in-position', {
+    inverse: 'position',
+  })
+  agentsInPosition;
 }
--- a/app/models/administrative-unit.js
+++ b/app/models/administrative-unit.js
@@ -2,8 +2,23 @@ import { hasMany, belongsTo } from '@ember-data/model';
 import OrganizationModel from './organization';
 
 export default class AdministrativeUnitModel extends OrganizationModel {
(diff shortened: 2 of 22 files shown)


app/models/post.js
```diff
@@ -1,7 +1,18 @@
...
-  @belongsTo('role') generalRole;
-  @belongsTo('organization') organization;
-  @hasMany('agent-in-position', { inverse: 'position' }) agentsInPosition;
+  @belongsTo('role', {
+    inverse: null,
+  })
+  generalRole;
+
+  @belongsTo('organization', {
+    inverse: null,
```
Windvis, line 10: Not sure about this one, but it was already marked as null in the organization model:

https://github.com/lblod/frontend-contact-hub/pull/85/files#diff-fca73d59c3b45630be23c80f2f7b814ed4a897bbc0c7926a38077df566f17b67L12

anitacaron:
@gcapurso, Sam also noticed the inverse problem and fixed it in this PR. Could you undo your changes, please?

app/models/post.js
anitacaron: Yeah, I think it should be linked to positions. It would be better for later.

Windvis: ok, so I link those 2 together?

anitacaron: The inverse here has to be the same as in the domain.json?

Windvis: Not sure about that. I still don't understand inverses in the linked data world :smile:

Windvis: I can't link this to positions without doing the inverse there I think.

anitacaron: Are you sure these changes didn't break anything? 😬

Windvis: No :grimacing: I clicked through some screens, but it's hard to test everything.

Windvis: Anything specific I should take a look at?

anitacaron (approved):

PR #88, opened 2021-07-30
Chore/people data loading improvements

Windvis:


app/routes/people/index.js
```diff
@@ -20,12 +20,10 @@ export default class PeopleIndexRoute extends Route {
   @dropTask({ cancelOn: 'deactivate' })
   *loadPeopleTask(params) {
     let query = {
-      // Includes slow down the API response so we disable them for now.
-      // include: [
-      //   'mandatories.status',
-      //   'mandatories.mandate.governing-body.is-time-specialization-of.administrative-unit',
-      //   'mandatories.mandate.role-board',
-      // ].join(),
+      include: [
```
Windvis, line 23: Not sure if this is a good idea, the requests will be cached but the first time it happens it will still be really slow.

Each filter value also counts as a unique request so every search will probably be slow.

Windvis: I'll remove this commit for now.

anitacaron (approved):

PR #93, opened 2021-08-03
Add minister position details page

gcapurso:


app/templates/people/person/positions/minister/index.hbs
```diff
@@ -0,0 +1,130 @@
+<div class="au-c-body-container au-c-body-container--scroll">
+  <div class="au-o-box au-o-flow au-o-flow--large">
+    <div class="au-o-flow au-o-flow--tiny">
```
Windvis, line 3: This structure can be replaced by the `PageHeader` component.

app/templates/people/person/positions/minister/index.hbs
```diff
@@ -0,0 +1,130 @@
...
+      <AuHeading @level="2" @skin="3" class="au-u-margin-top-tiny">
+        {{@model.person.givenName}}
+        {{@model.person.familyName}}
+      </AuHeading>
+    </div>
+
+    <DataCard>
+      <:title>Positie</:title>
+      <:card as |Card|>
+        <Card.Grid as |Item|>
```
Windvis, line 30: I would use the Card.Colums version instead, that gives you more control over the placement of the data.

app/templates/people/person/positions/minister/index.hbs
```diff
@@ -0,0 +1,130 @@
...
+          <Item>
+            {{! TODO: Make it variable to also have Bedienaar}}
```
Windvis, line 32: Since this page is now dedicated to minister positions, I think this comment can simply be removed?

app/templates/people/person/positions/minister/index.hbs
```diff
@@ -0,0 +1,130 @@
...
+              as |administrativeUnit|
+              }}
+              <AuLink @route="administrative-units.administrative-unit" @model={{administrativeUnit.id}}>
+                {{administrativeUnit.name}}
+              </AuLink>
+              {{/let}}
+            </:content>
+          </Item>
+          <Item>
+            <:label>Bestuursfunctie</:label>
```
Windvis, line 50: I think this should be `Rol` for the minister version.

app/templates/people/person/positions/minister/index.hbs
```diff
@@ -0,0 +1,130 @@
...
+          {{#if @model.minister.contacts.firstObject.email}}
+          <Item>
+            <:label>Email</:label>
+            <:content>
+              {{@model.minister.contacts.firstObject.email}}
+            </:content>
+          </Item>
+          {{/if}}
+          {{#if @model.minister.contacts.firstObject.telephone}}
+          {{! TODO: Identify if it is a landline or mobile number }}
```
Windvis, line 71: This comment can be removed.

Windvis (changes requested):
First pass review, I'll test your branch after lunch but wanted to give the current feedback already :smile:.

app/templates/people/person/positions/minister/index.hbs
Windvis: (Be sure to also remove this in the mandatory version. I can't leave a comment since the PR doesn't change that file.)

app/templates/people/person/positions/minister/index.hbs
```diff
@@ -0,0 +1,130 @@
...
+          @model.minister.contacts.firstObject.contactAddress.fullAddress
+          }}
+          <Item>
+            <:label>Adres</:label>
+            <:content>
+              {{@model.minister.contacts.firstObject.contactAddress.fullAddress}}
+            </:content>
+          </Item>
+          {{/if}}
+          {{#if (get (array-index @model.minister.contacts 1) 'telephone')}}
```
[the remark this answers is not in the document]
gcapurso: Apparently it is not working. I tried bot {{#if (get (get @model.minister.contacts 1) "telephone")}} and {{#if (get @model.minister.contacts "1.telephone")}}.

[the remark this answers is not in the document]
Windvis: Hmm, maybe the ember data specific array isn't supported. It definitely works for a normal array.

In that case I would use ember-composable-helpers' [`object-at`](https://github.com/DockYard/ember-composable-helpers#object-at) helper instead.

app/routes/people/person/positions/minister.js
```diff
@@ -0,0 +1,24 @@
...
+  async model() {
+    let person = this.modelFor('people.person');
+
+    let { positionId: ministerId } = this.paramsFor(
+      'people.person.positions.minister'
+    );
+
+    let minister = this.store.findRecord('minister', ministerId, {
+      reload: true,
+      include: ['contacts', 'minister-position'].join(),
```
Windvis, line 16: It seems there are still some async calls happening, so I would include those relationships as well:
image

Windvis:
It seems the `people.person.positions.mandatory.index` page shows an error due to this line:
https://github.com/lblod/frontend-contact-hub/blob/846f2a96dd82028a186958c2e5b9ca42cf16de98/app/templates/people/person/positions/mandatory/index.hbs#L23

That route no longer exists, so the link component fails on it.

app/templates/people/person/positions/minister/index.hbs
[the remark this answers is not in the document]
gcapurso: Yeah, this is the right way, thank you!

Windvis:
I also think the ministers page of a administrative unit should link to the minister position details page.

So here: https://github.com/lblod/frontend-contact-hub/blob/c764e88332b544d48a8eda8dc34575953fde2a15/app/templates/administrative-units/administrative-unit/ministers/index.hbs#L42-L44

app/routes/people/person/positions/minister.js
gcapurso: Thank you, I just learned to check the network tab to discover the relationships that should be included!

app/routes/people/person/positions/mandatory.js
```diff
@@ -1,14 +1,14 @@
...
 import { inject as service } from '@ember/service';
 
-export default class PeoplePersonPositionsPositionRoute extends Route {
+export default class PeoplePersonPositionsMandatoryRoute extends Route {
   @service store;
 
   async model() {
     let person = this.modelFor('people.person');
 
     let { positionId: mandatoryId } = this.paramsFor(
```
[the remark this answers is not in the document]
gcapurso: I'm not sure I understood it, how it should be? because than I need the positionId.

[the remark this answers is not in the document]
Windvis: Yea, so the mandatory route has a `positionId` dynamic segment. This meant that the model hook of that route will receive the value for that as the first argument to the model method.

Similar to this for the person route:
https://github.com/lblod/frontend-contact-hub/blob/aaa5c7f070120a5b97799ec50c8702530e5eed30/app/routes/people/person.js#L4

So instead of using `this.paramsFor` you can get the id from that first argument instead:

`async model({ positionId: mandatoryId }) {...`

paramsFor is only needed if you want to retrieve params for a parent route.

Windvis (changes requested):

PR #95, opened 2021-08-04
Feature/mock login

nbittich:


app/templates/mock-login.hbs
```diff
@@ -0,0 +1,39 @@
+<div class="au-c-main-container" itemprop="mainContentOfPage">
+    <div class="au-c-main-container__content au-c-main-container__content--scroll">
+        <main id="content">
+            <div class="au-o-region-large">
+                <div class="au-o-layout">
+                    {{#if this.session.isAuthenticated}}
+                    Logged in
+                    <MockLogout />
+                    {{else}}
```
Windvis, line 9: This part isn't needed since a logged in user will be redirected to the index page already.

Windvis (changes requested):

Windvis (changes requested):

Windvis:
I think the FastBoot support part is missing at the moment: https://github.com/simplabs/ember-simple-auth#fastboot

Windvis:
Rebased on master and removed the cookie store since we don't need to support FastBoot anymore.

PR #96, opened 2021-08-04
Add wording change email and type

gcapurso:
https://binnenland.atlassian.net/browse/OP-438
https://binnenland.atlassian.net/browse/OP-435

### DIFF
--- a/app/templates/administrative-units/administrative-unit/ministers/new.hbs
+++ b/app/templates/administrative-units/administrative-unit/ministers/new.hbs
@@ -84,7 +84,7 @@
                   <a
                     href="mailto:{{config "contactEmail"}}"
                     class="au-c-link"
-                  >{{config "contactEmail"}}e</a>.
+                  >{{config "contactEmail"}}</a>.
                 </p>
               </TableMessage>
             {{else}}
--- a/app/templates/administrative-units/administrative-unit/ministers/new.hbs
+++ b/app/templates/administrative-units/administrative-unit/ministers/new.hbs
@@ -82,9 +82,9 @@
                   of probeer een andere zoekopdracht. Indien een persoon
                   ontbreekt, meldt dit aan
                   <a
-                    href="mailto:organisaties.abb@vlaanderen.be"
+                    href="mailto:{{config "contactEmail"}}"
                     class="au-c-link"
-                  >organisaties.abb@vlaanderen.be</a>.
+                  >{{config "contactEmail"}}e</a>.
                 </p>
               </TableMessage>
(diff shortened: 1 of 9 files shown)


app/templates/administrative-units/administrative-unit/ministers/new.hbs
```diff
@@ -82,9 +82,9 @@
                   of probeer een andere zoekopdracht. Indien een persoon
                   ontbreekt, meldt dit aan
                   <a
-                    href="mailto:TODO"
+                    href="mailto:organisaties.abb@vlaanderen.be"
```
Windvis, line 85: Since this is used in multiple pages I think it might be better to store this in the environment.js file and use the config helper to retrieve it.

Quick summary:
1. add a `contactEmail` or `helpDeskEmail` to the environment.js file, similar to this:
https://github.com/lblod/frontend-contact-hub/blob/6c2b2c90c6fbfb67328fcb6db0c903582df291eb/config/environment.js#L5
2. replace all the hardcoded emails with the config that can be retrieved with the config helper like this:
https://github.com/lblod/frontend-contact-hub/blob/6c2b2c90c6fbfb67328fcb6db0c903582df291eb/app/templates/application.hbs#L6
3. I think this also needs to be changed in the contact page:
https://github.com/lblod/frontend-contact-hub/blob/6c2b2c90c6fbfb67328fcb6db0c903582df291eb/app/controllers/contact.js#L4
This can be removed and replaced with a config import:

```js
import config from 'frontend-contact-hub/config/environment';

// ...

contactEmail = config.contactEmail;
```

Windvis (changes requested):

app/templates/administrative-units/administrative-unit/ministers/new.hbs
```diff
@@ -82,9 +82,9 @@
                   of probeer een andere zoekopdracht. Indien een persoon
                   ontbreekt, meldt dit aan
                   <a
-                    href="mailto:organisaties.abb@vlaanderen.be"
+                    href="mailto:{{config "contactEmail"}}"
                     class="au-c-link"
-                  >organisaties.abb@vlaanderen.be</a>.
+                  >{{config "contactEmail"}}e</a>.
```
Windvis, line 87: There's a ninja `e` in there :smile:.

gcapurso: Or maybe a mosquito on your monitor :smile_cat:

Windvis (approved):

PR #99, opened 2021-08-05
Replace hardcoded user information with real data

Windvis:


app/components/app-header.js
```diff
@@ -0,0 +1,11 @@
+import Component from '@glimmer/component';
+import { inject as service } from '@ember/service';
+
+export default class AppHeaderComponent extends Component {
+  @service currentSession;
+  @service session;
+
+  logout = () => {
```
Windvis, line 8: woops forget to replace this with an action.

gcapurso (approved):

PR #102, opened 2021-08-09
Fix/responsive sidebar

Dietr:
- makes the sidebar full width on smaller widths
- layout the filters in a 2x2 grid on smaller widths

### DIFF
--- a/app/templates/index.hbs
+++ b/app/templates/index.hbs
@@ -3,7 +3,7 @@
     <div class="au-c-sidebar">
       <div class="au-c-sidebar__content">
         <div class="au-c-sidebar__header">
-          <p class="au-u-h5 au-u-medium">Menu modules</p>
+          <h6 class="au-u-h6 au-u-medium">Menu modules</h6>
         </div>
         <ul class="au-c-list-navigation">
           <li class="au-c-list-navigation__item">
--- a/app/templates/people/index.hbs
+++ b/app/templates/people/index.hbs
@@ -2,8 +2,8 @@
   <:sidebar>
     <div class="au-c-sidebar">
       <div class="au-c-sidebar__content au-o-box">
-        <form {{on "submit" this.search}} class="au-c-form">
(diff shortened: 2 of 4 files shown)


app/templates/index.hbs
```diff
@@ -3,7 +3,7 @@
     <div class="au-c-sidebar">
       <div class="au-c-sidebar__content">
         <div class="au-c-sidebar__header">
-          <p class="au-u-h5 au-u-medium">Menu modules</p>
+          <h6 class="au-u-h6 au-u-medium">Menu modules</h6>
```
Windvis, line 6: I don't think it needs to be a h6 because it's styled like one?

app/templates/people/index.hbs
```diff
@@ -2,39 +2,41 @@
   <:sidebar>
     <div class="au-c-sidebar">
       <div class="au-c-sidebar__content au-o-box">
-        <form {{on "submit" this.search}} class="au-c-form">
-          <p>
+        <form {{on "submit" this.search}} class="au-o-grid au-o-grid--small">
+          <p class="au-o-grid__item">
```
Windvis, line 6: I think it might be better to show the name inputs next to each other instead of last name and organization name?

Windvis (changes requested):

app/templates/index.hbs
Dietr: Can be another h* but as it's a title it should be marked up as a title element. A p tag does not seem correct here.

Windvis: Yea, fair enough. I usually don't use h elements if the thing shouldn't be visible in the fictive "outline" which is why we used a p tag here.

I think using a h6 is a weird choice as well :smile: since it just seems completely dependent on the styling. I'm no expert on semantics and the spec isn't very clear either so I'll drop it.

Windvis (approved):

PR #103, opened 2021-08-09
add sorting on status

Dietr:
- added status sorting on administrative unit

Note: status sorting on person still needs to be done but I need some help with that as the filter is not yet available.


Windvis:
The people sort is tricky since it's not an attribute of the records. 

We use this logic which can't really be mimicked by sorting:
https://github.com/lblod/frontend-contact-hub/blob/d9882ad58e387ac8ed54064c255af178086b556a/app/components/mandatory-status.js#L8-L13

Maybe sorting on the endDate might have the effect we want by accident?

PR #101, opened 2021-08-06
governing body naming refactoring and breadcrumb bug fix

Windvis:
I noticed the governing body breadcrumb wasn't working properly and while trying to figure out the issue I noticed how confusing the naming was, so I tried fixing that as well :grimacing:.

PR #104, opened 2021-08-10
Fix the mandatory edit page

Windvis:
- rename regression caused by the previous refactor (#101)
- link the address records correctly


anitacaron:
ops, thanks for these fix 😬

PR #105, opened 2021-08-10
Feature/acm login rebase

nbittich:


config/environment.js
```diff
@@ -25,6 +25,19 @@ module.exports = function (environment) {
       // Here you can pass flags/options to your application instance
       // when it is created
     },
+    torii: {
+      disableRedirectInitializer: true,
+      providers: {
+        'acmidm-oauth2': {
+          apiKey: '677af572-2e2f-4e61-ad38-036723bb314c',
```
Windvis, line 32: Do these values need to be checked in? I think just using the placeholders here would be better?

That seems to be how GN does it:
https://github.com/lblod/frontend-gelinkt-notuleren/blob/e344376bda3b8aa276ec29407fb7a5842b3160cc/config/environment.js#L56-L67

config/environment.js
```diff
@@ -51,5 +64,17 @@ module.exports = function (environment) {
     // here you can enable a production-specific feature
   }
 
+  if (environment === 'production') {
+    // TODO set the proper value for production
```
Windvis, line 68: This TODO comment isn't needed right, since it will be configured on the server?

Windvis (changes requested):
The button in the header needs to start the login flow as well:
https://github.com/lblod/frontend-contact-hub/blob/7ad7f711f513f9d738a1184b91f4dea36b538076/app/components/app-header.hbs#L21-L28

The problem is that the `<AcmidmLogin>` is hardcoded to show a primary button, which won't work in the header. Other apps create a compact version of the component instead, but I don't like that. I think the real solution would be to make the components more flexible instead. https://github.com/lblod/ember-acmidm-login/pull/4#discussion_r685131726

Maybe we can simply hide the header button for now? I can do the work in the acmidm addon but I would like to discuss it with Niels first.

config/environment.js
nbittich: yes probably, I did it this way  so we can test using ember serve, but anyway i don't think that will work (thus using mock login in that case). i'll change it

nbittich:
> The button in the header needs to start the login flow as well:
> https://github.com/lblod/frontend-contact-hub/blob/7ad7f711f513f9d738a1184b91f4dea36b538076/app/components/app-header.hbs#L21-L28
> 
> The problem is that the `<AcmidmLogin>` is hardcoded to show a primary button, which won't work in the header. Other apps create a compact version of the component instead, but I don't like that. I think the real solution would be to make the components more flexible instead. [lblod/ember-acmidm-login#4 (comment)](https://github.com/lblod/ember-acmidm-login/pull/4#discussion_r685131726)
> 
> Maybe we can simply hide the header button for now? I can do the work in the acmidm addon but I would like to discuss it with Niels first.

probably better to hide it for now and check it later once Niels get back from holidays; maybe we can create a jira ticket for it ? also when Niels get back from holidays we could also eventually merge his PR and release a new version of the addon

PR #106, opened 2021-08-10
Show commit hash in dev environments

Windvis:
This will show the commit hash (which links to github :sunglasses:) instead of the version in the dev environment. This hopefully will make things less confusing :smile:. 

image


Windvis:
I tested building the image locally which seemed to work, so hopefully it still works in Drone :grimacing:.

nbittich (approved):

PR #118, opened 2021-08-16
Create Position Status component

anitacaron:
Create a more generic component would be better, like `PositionStatus`. What do you think?


app/components/minister-status.js
```diff
@@ -0,0 +1,22 @@
...
+    if (!endDate) {
+      // No end date set, so the mandatory is still active
+      return true;
+    } else {
+      return isDateInTheFuture(endDate);
+    }
+  }
+}
+
+function isDateInTheFuture(date) {
```
Windvis, line 17: I think we use this elsewhere already? Maybe we can export this from a `app/utils/date.js` file or something?

Windvis (approved):

Windvis:
> Create a more generic component would be better, like `PositionStatus`. What do you think?

Hmm, not sure. How would that work then? Check both the minister and mandatory status and return active when one of them is? Where do we need this exactly?

anitacaron:
For both, we only check the end date, we can pass this to the component, and that's it.

app/components/minister-status.js
anitacaron: Yes, that's why I'm suggesting creating a generic component and pass only the end date.

Windvis: Ok! Didn't know where we used the other one. Yea in that case renaming it to PositionStatus makes sense!

Windvis (approved):
Nice!

PR #120, opened 2021-08-17
404

Dietr:
Added 404 route and template

### DIFF
--- a/app/templates/route-not-found.hbs
+++ b/app/templates/route-not-found.hbs
@@ -2,7 +2,7 @@
 
 <div class="au-c-body-container au-c-body-container--scroll">
   <div class="au-o-region-large">
-    <div class=" au-o-layout">
+    <div class="au-o-layout">
       <AuContent @skin="large">
         <h1>We kunnen de pagina niet vinden</h1>
         <p>Er is iets fout gelopen. De link die je bezocht, bestaat niet.</p>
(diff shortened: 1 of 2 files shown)


app/templates/route-not-found.hbs
```diff
@@ -0,0 +1,12 @@
+{{page-title "404 - Pagina niet gevonden"}}
```
Windvis, line 1: Is the `404` needed in the title? Seems like a technical detail that the user doesn't really need to know?

app/templates/route-not-found.hbs
```diff
@@ -0,0 +1,12 @@
...
+
+<div class="au-c-body-container au-c-body-container--scroll">
+  <div class="au-o-region-large">
+    <div class=" au-o-layout">
+      <AuContent @skin="large">
+        <h1>We kunnen de pagina niet vinden</h1>
```
Windvis, line 7: Just a question. When should the AuContent component be used instead the AuHeading one for example? I assumed it was needed to style dynamic / user generated content but since it's used here I don't think it's that simple?

Dietr: Generally I use the AuHeading component for titles in the interface. I use the AuContent component when there is user-generated content or grouped text (titles and paragraphs). Benefit is that the AuContent component also has spacing set which helps when there is a combo of titles and paragraphs.

PR #127, opened 2021-08-19
Governing body details page improvements

Windvis:
- show all the mandatories attached to each role
- fix the count in the pagination
- add sorting to the tables

### DIFF
--- a/app/routes/administrative-units/administrative-unit/governing-bodies/governing-body/index.js
+++ b/app/routes/administrative-units/administrative-unit/governing-bodies/governing-body/index.js
@@ -17,7 +17,11 @@ const OTHER_ROLES = [
 export default class AdministrativeUnitsAdministrativeUnitGoverningBodiesGoverningBodyIndexRoute extends Route {
   @service store;
 
-  async model() {
+  queryParams = {
+    sort: { refreshModel: true },
+  };
+
+  async model({ sort }) {
     let { administrativeUnit, governingBodyClassification, governingBody } =
       this.modelFor(
         'administrative-units.administrative-unit.governing-bodies.governing-body'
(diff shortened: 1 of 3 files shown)


app/routes/administrative-units/administrative-unit/governing-bodies/governing-body/index.js
```diff
@@ -1,29 +1,47 @@
...
   async model() {
     let { administrativeUnit, governingBodyClassification, governingBody } =
       this.modelFor(
         'administrative-units.administrative-unit.governing-bodies.governing-body'
       );
 
-    let mandates = await governingBody.mandates;
-    let memberMandates = mandates.filter((mandate) => {
-      return mandate.get('roleBoard.id') === MEMBER_ROLE_ID;
+    let governingBodyMandatories = await this.store.query('worship-mandatory', {
```
Windvis, line 15: Hmm, the current version will be limited to 20 records since that's the default. And pagination won't work as expected since we split the array manually. Ideally we could just do 2 fetches where the filtering is done server side, but `NOT` checks aren't supported last time I checked.

anitacaron: Can't you edit the page size in the query?

anitacaron: From what I've seen in the data, 20 records will be enough anyway, but it's good to keep this in mind or add a comment.

Windvis: Yea, I was trying to make the `not` version work with a `:gt:`  and `:lt:` filter hack. but that doesn't seem to work for ids.

Windvis: A `not` filter would be real handy though.

Windvis: I solved it by including all the ids (besides the lid one) :smile:, best hacky workaround I could find.

Windvis:
As discussed, I've disabled the sorting for the date fields for now.

anitacaron (approved):

PR #128, opened 2021-08-23
Implement the local involvements edit and

Windvis:


app/templates/administrative-units/administrative-unit/local-involvements/edit.hbs
```diff
@@ -0,0 +1,71 @@
+{{page-title "Bewerk"}}
+
```
Windvis, line 2: I forgot the wrapping div with the scroll class which is probably needed on this page.

anitacaron: Ok

PR #130, opened 2021-08-23
Temporarily hide links to unused pages

mietcls:
To make sure we don't have to create new pages explaining why these pages do not exist just yet. We can show them again once the pages have been implemented – perhaps add a "new" label or a nice onboarding experience to the new feature.


Windvis (approved):
Thanks!

PR #132, opened 2021-08-24
Alert for new entitites and missing functionalities

Dietr:
Includes:
- Personen toevoegen
- Contactgegevens bewerken
- Posities toevoegen & bewerken
- Bestuursorganen
- Vestigingen toevoegen
- Betrokken lokale besturen
- Bedienaren

### DIFF
--- a/app/components/person-search.hbs
+++ b/app/components/person-search.hbs
@@ -96,7 +96,9 @@
     </AuDataTable>
   </div>
 {{else}}
-  <p class="au-u-margin au-u-margin-top-huge">
-    Zoek een reeds gekende persoon in de databank van het loket op naam.
-  </p>
+  <div class="au-o-box">
+    <AuAlert @alertIcon="info-circle" @alertTitle="Zoek een reeds gekende persoon in de databank van het loket op naam." @alertskin="info" @alertsize="small" class="au-u-margin-bottom-none">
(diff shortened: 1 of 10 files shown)


app/components/person-search.hbs
```diff
@@ -96,7 +96,9 @@
     </AuDataTable>
   </div>
 {{else}}
-  <p class="au-u-margin au-u-margin-top-huge">
-    Zoek een reeds gekende persoon in de databank van het loket op naam.
-  </p>
+  <div class="au-o-box">
+    <AuAlert @alertIcon="info-circle" @alertTitle="Zoek een reeds gekende persoon in de databank van het loket op naam." @alertskin="info" @alertsize="small" class="au-u-margin-bottom-none">
```
Windvis, line 100: I'm not sure if this needed to be changed to an alert. The JIRA issue makes it sound like the text should have been appended to the existing message.

Dietr: It's more consistent to make this in to an alert. The message contains similar information as the other alerts so makes sense to design it the same way.

Windvis: The message says that you can search people by name. The not being able to add a person is just extra info, so I don't think it's the same use-case, but fair enough.

PR #145, opened 2021-08-31
Make sure the mandatory tables show the correct roles

Windvis:


Windvis:
The discussion in the issue needs to be resolved before this can be considered done.

anitacaron (approved):

PR #150, opened 2021-09-02
Change related organization based on the admin unit type

anitacaron:


app/templates/administrative-units/new.hbs
```diff
@@ -243,6 +235,19 @@
       <:card as |Card|>
         <Card.Columns>
           <:left as |Item|>
+          {{#if this.isNewCentralWorshipService}}
+            <Item>
+              <:label>Bestuur van de eredienst</:label>
+              <:content>
+                <WorshipServiceSelect
+                  @selected={{this.model.administrativeUnit.subOrganizations}}
+                  @onChange={{fn
```
Windvis, line 244: Selecting doesn't seem to work yet. The relationship expects an array instead of a single record. So either the response needs to be wrapped in an array, or the select needs to be converted to a multi select.

Windvis (approved):

PR #152, opened 2021-09-03
Edit minister page

gcapurso:
https://www.figma.com/file/P8aIfSN0f1CctxBGYdeyKe/ERE-v1?node-id=621%3A5504

### DIFF
--- a/app/controllers/people/person/positions/minister/edit.js
+++ b/app/controllers/people/person/positions/minister/edit.js
@@ -2,6 +2,7 @@ import Controller from '@ember/controller';
 import { inject as service } from '@ember/service';
 import { dropTask } from 'ember-concurrency';
 import { tracked } from '@glimmer/tracking';
+import { combineFullAddress } from 'frontend-contact-hub/models/address';
 
 const FINANCING_CODE = {
   SELF_FINANCED: '997073905f839ac6bafe92b76050ab0b',
@@ -16,6 +17,16 @@ export default class PeoplePersonPositionsMinisterEditController extends Control
   @dropTask
   *save(event) {
     event.preventDefault();
+
+    let contacts = yield this.model.minister.contacts;
+    let address = yield contacts.firstObject.contactAddress;
+
+    if (address.isNew || address.hasDirtyAttributes) {
+      address.fullAddress = combineFullAddress(address);
+      yield address.save();
+    }
+
+    yield contacts.firstObject.save();
 
     let financingCodeId = this.willReceiveFinancing
       ? FINANCING_CODE.FOD_FINANCED
--- a/app/controllers/people/person/positions/minister/edit.js
+++ b/app/controllers/people/person/positions/minister/edit.js
@@ -1,5 +1,6 @@
 import Controller from '@ember/controller';
 import { inject as service } from '@ember/service';
+import { action } from '@ember/object';
 import { dropTask } from 'ember-concurrency';
 import { tracked } from '@glimmer/tracking';
 import { combineFullAddress } from 'frontend-contact-hub/models/address';
@@ -12,16 +13,30 @@ const FINANCING_CODE = {
 export default class PeoplePersonPositionsMinisterEditController extends Controller {
   @service router;
   @tracked willReceiveFinancing;
-  @tracked isCurrentPositionCheck;
+  @tracked isCurrentPosition;
+  @tracked redirectUrl;
 
-  get isCurrentPosition() {
-    return this.model.minister.agentEndDate === null;
-  }
+  queryParams = ['redirectUrl'];
 
   setup() {
     this.willReceiveFinancing =
       this.model.minister.financing.get('id') === FINANCING_CODE.FOD_FINANCED;
-    this.isCurrentPositionCheck = this.isCurrentPosition;
+    this.isCurrentPosition = !this.model.minister.agentEndDate;
+  }
+
+  @action
+  async clearOnCheck() {
+    this.model.minister.agentEndDate = undefined;
+    this.isCurrentPosition = true;
+  }
+
+  @action
+  cancel() {
+    if (this.redirectUrl) {
+      this.router.transitionTo(this.redirectUrl);
+    } else {
+      this.router.transitionTo('people.person.positions.minister');
+    }
   }
 
   @dropTask
@@ -53,9 +68,18 @@ export default class PeoplePersonPositionsMinisterEditController extends Control
     this.model.minister.financing = financing;
     yield this.model.minister.save();
 
-    this.router.transitionTo(
-      'people.person.positions.minister',
-      this.model.minister.id
-    );
+    if (this.redirectUrl) {
+      // When passing a url the query params are ignored so we add the person id manually for now
+      this.router.transitionTo(this.redirectUrl);
+    } else {
+      this.router.transitionTo(
+        'people.person.positions.minister',
+        this.model.minister.id
+      );
+    }
+  }
+
+  reset() {
+    this.redirectUrl = null;
   }
 }
--- a/app/controllers/people/person/positions/minister/edit.js
+++ b/app/controllers/people/person/positions/minister/edit.js
@@ -1,0 +1,40 @@
+import Controller from '@ember/controller';
+import { inject as service } from '@ember/service';
+import { dropTask } from 'ember-concurrency';
+import { tracked } from '@glimmer/tracking';
+
+const FINANCING_CODE = {
+  SELF_FINANCED: '997073905f839ac6bafe92b76050ab0b',
+  FOD_FINANCED: '9d6f49b3d923b437ec3a91e8b5fa6885',
+};
+
+export default class PeoplePersonPositionsMinisterEditController extends Controller {
+  @service router;
+
+  @tracked willReceiveFinancing;
+
+  @dropTask
+  *save(event) {
+    event.preventDefault();
+
+    let financingCodeId = this.willReceiveFinancing
+      ? FINANCING_CODE.FOD_FINANCED
+      : FINANCING_CODE.SELF_FINANCED;
+
+    let financing = yield this.store.findRecord(
+      'financing-code',
+      financingCodeId,
+      {
+        backgroundReload: false,
+      }
+    );
+
+    this.model.minister.financing = financing;
+    yield this.model.minister.save();
+
+    this.router.transitionTo(
+      'people.person.positions.minister',
+      this.model.minister.id
+    );
+  }
+}
--- a/app/controllers/people/person/positions/minister/edit.js
+++ b/app/controllers/people/person/positions/minister/edit.js
@@ -24,19 +24,25 @@ export default class PeoplePersonPositionsMinisterEditController extends Control
     this.isCurrentPosition = !this.model.minister.agentEndDate;
   }
 
-  @action
-  async clearOnCheck() {
-    this.model.minister.agentEndDate = undefined;
-    this.isCurrentPosition = true;
-  }
-
-  @action
-  cancel() {
+  handleTransitionTo() {
     if (this.redirectUrl) {
       this.router.transitionTo(this.redirectUrl);
     } else {
       this.router.transitionTo('people.person.positions.minister');
     }
+  }
+
+  @action
+  async handleIsCurrentPositionChange() {
+    this.isCurrentPosition = !this.isCurrentPosition;
+    if (this.isCurrentPosition) {
+      this.model.minister.agentEndDate = undefined;
+    }
+  }
+
+  @action
+  cancel() {
+    this.handleTransitionTo();
   }
 
   @dropTask
@@ -68,15 +74,7 @@ export default class PeoplePersonPositionsMinisterEditController extends Control
     this.model.minister.financing = financing;
     yield this.model.minister.save();
 
-    if (this.redirectUrl) {
-      // When passing a url the query params are ignored so we add the person id manually for now
-      this.router.transitionTo(this.redirectUrl);
-    } else {
-      this.router.transitionTo(
-        'people.person.positions.minister',
-        this.model.minister.id
-      );
-    }
+    this.handleTransitionTo();
   }
 
   reset() {
--- a/app/controllers/people/person/positions/minister/edit.js
+++ b/app/controllers/people/person/positions/minister/edit.js
@@ -33,7 +33,7 @@ export default class PeoplePersonPositionsMinisterEditController extends Control
   }
 
   @action
-  async handleIsCurrentPositionChange() {
+  handleIsCurrentPositionChange() {
     this.isCurrentPosition = !this.isCurrentPosition;
     if (this.isCurrentPosition) {
       this.model.minister.agentEndDate = undefined;
--- a/app/controllers/people/person/positions/minister/edit.js
+++ b/app/controllers/people/person/positions/minister/edit.js
@@ -11,8 +11,18 @@ const FINANCING_CODE = {
 
 export default class PeoplePersonPositionsMinisterEditController extends Controller {
   @service router;
-
   @tracked willReceiveFinancing;
+  @tracked isCurrentPositionCheck;
+
+  get isCurrentPosition() {
+    return this.model.minister.agentEndDate === null;
+  }
+
+  setup() {
+    this.willReceiveFinancing =
+      this.model.minister.financing.get('id') === FINANCING_CODE.FOD_FINANCED;
+    this.isCurrentPositionCheck = this.isCurrentPosition;
+  }
 
   @dropTask
   *save(event) {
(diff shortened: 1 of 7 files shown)


app/templates/people/person/positions/minister/edit.hbs
```diff
@@ -0,0 +1,167 @@
...
+                            <MinisterPositionFunctionSelect
+                                @onChange={{fn (mut this.model.minister.ministerPosition.function)}}
+                                @selected={{this.model.minister.ministerPosition.function}}
+                            />
+                            </:content>
+                        </Item>
+                        <Item>
+                            <:label>Type</:label>
+                            <:content>
+                                <PowerSelect
```
Windvis, line 42: I would just use plain text instead of a disabled power select.

app/templates/people/person/positions/minister/edit.hbs
```diff
@@ -0,0 +1,167 @@
...
+                        </:content>
+                    </Item>
+                    </:left>
+                </Card.Columns>
+                </:card>
+            </EditCard>
+
+            {{! TODO: add "Voorwaarde" section once file uploads are supported }}
+
+            <div>
```
Windvis, line 150: Dieter opened a PR to make the buttons more consistent, so I would use the same setup here.

A [ButtonGroup component](https://github.com/lblod/frontend-contact-hub/pull/154/files#diff-fc99dcdfb5b02a9c536c332ea5f414ef9cbbf0fbd4e70b37640f83d6a86694c3R253) should be used here.

app/templates/people/person/positions/minister/edit.hbs
```diff
@@ -0,0 +1,167 @@
...
+                <AuButton
+                    @iconAlignment="left"
+                    @loading={{this.save.isRunning}}
+                    @disabled={{this.save.isRunning}}
+                    class="au-u-margin-right-small"
```
Windvis, line 155: This class shouldn't be needed when the ButtonGroup component is used.

app/templates/people/person/positions/minister/edit.hbs
```diff
@@ -0,0 +1,167 @@
...
+                    type="submit"
+                >
+                    Opslaan
+                </AuButton>
+                <AuLink @route="people.person.positions.minister">
```
Windvis, line 160: This should be [styled as a button](https://github.com/lblod/frontend-contact-hub/pull/154/files#diff-fc99dcdfb5b02a9c536c332ea5f414ef9cbbf0fbd4e70b37640f83d6a86694c3R262-R264).

app/templates/people/person/positions/minister/edit.hbs
```diff
@@ -0,0 +1,167 @@
...
+                <:card as |Card|>
+                <Card.Columns>
+                    <:left as |Item|>
+                        <Item>
+                            <:label>Rol</:label>
+                            <:content>
+                            <MinisterPositionFunctionSelect
+                                @onChange={{fn (mut this.model.minister.ministerPosition.function)}}
+                                @selected={{this.model.minister.ministerPosition.function}}
+                            />
```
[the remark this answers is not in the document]
Windvis: Confirmed by Yassin in 310. The first card should not be editable, similar to the mandatory edit page.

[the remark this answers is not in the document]
gcapurso: Perfect, so I'll put labels with text instead PowerSelect

app/controllers/people/person/positions/minister/edit.js
```diff
@@ -0,0 +1,51 @@
...
+
+const FINANCING_CODE = {
+  SELF_FINANCED: '997073905f839ac6bafe92b76050ab0b',
+  FOD_FINANCED: '9d6f49b3d923b437ec3a91e8b5fa6885',
+};
+
+export default class PeoplePersonPositionsMinisterEditController extends Controller {
+  @service router;
+
+  @tracked willReceiveFinancing;
```
Windvis, line 15: At the moment this will always check the "no" radio button because `undefined` is a falsy value.
Th
is needs to be set to a value when the page is loaded. So adding a "setup" method which compares the `minister.financing.id` to the `FINANCING_CODE.FOD_FINANCED` might be a good enough solution for now.

app/controllers/people/person/positions/minister/edit.js
```diff
@@ -0,0 +1,51 @@
...
+
+  @dropTask
+  *save(event) {
+    event.preventDefault();
+
+    let contacts = yield this.model.minister.contacts;
+    let address = yield contacts.firstObject.contactAddress;
+
+    if (address.isNew || address.hasDirtyAttributes) {
```
Windvis, line 24: We're editing an existing minister, so the isNew check will never be true so I would remove that.

app/templates/people/person/positions/minister/edit.hbs
```diff
@@ -0,0 +1,260 @@
+{{page-title "Bewerk"}}
+
+<div class="au-c-body-container au-c-body-container--scroll">
+    <div class="au-o-box au-o-flow au-o-flow--large">
+        <PageHeader>
+        <:title>
+            Positie:
+            {{@model.minister.ministerPosition.function.label}}
+            <br />
+            {{@model.minister.ministerPosition.worshipService.classification.label}}
```
Windvis, line 10: I don't think the classification label needs to be shown here? It looks a bit weird I think. Did you see it in a design or was it mentioned in a discussion somewhere?

app/templates/people/person/positions/minister/edit.hbs
```diff
@@ -0,0 +1,260 @@
...
+                        />
+                        </:content>
+                    </Item>
+                    </:left>
+                    <:right as |Item|>
+                    <Item @labelFor="end-date">
+                        <:label>Einddatum</:label>
+                        <:content>
+                        <Datepicker
+                            @disabled={{this.isCurrentPosition}}
```
Windvis, line 180: `this.isCurrentPosition` also needs to be determined when the page is visited. I assume it can be set to true if no `agentEndDate` is set.

Windvis (changes requested):

app/templates/people/person/positions/minister/edit.hbs
gcapurso: The minister index page shows it, that's why I "inherited" the classification. But in fact none of the two pages has it in Figma. We should probably change both of them?

app/controllers/people/person/positions/minister/edit.js
```diff
@@ -0,0 +1,61 @@
...
+  FOD_FINANCED: '9d6f49b3d923b437ec3a91e8b5fa6885',
+};
+
+export default class PeoplePersonPositionsMinisterEditController extends Controller {
+  @service router;
+  @tracked willReceiveFinancing;
+  @tracked isCurrentPositionCheck;
+
+  get isCurrentPosition() {
+    return this.model.minister.agentEndDate === null;
```
Windvis, line 18: This check is currently not working as it should. The value will be undefined instead of null when it's not set and `null !== undefined` so it will return false when it shouldn't.

`return !this.model.minister.agentEndDate;` should work and covers both the case were it isn't set yet, or when the data is explicitly cleared again.

app/templates/people/person/positions/minister/edit.hbs
```diff
@@ -0,0 +1,259 @@
...
+                <Card.Columns>
+                <:left as |Item|>
+                    {{#if this.model.minister.contacts.firstObject.contactAddress.fullAddress}}
+                    <Item>
+                        <:label>Full address</:label>
+                        <:content>
+                        {{this.model.minister.contacts.firstObject.contactAddress.fullAddress}}
+                        </:content>
+                    </Item>
+                    {{/if}}
```
Windvis, line 64: I think this can be removed. The new data should have proper address information instead of only the fullAddress attribute.

app/templates/people/person/positions/minister/edit.hbs
```diff
@@ -0,0 +1,259 @@
...
+                        </:content>
+                    </Item>
+                    <Item @labelFor="end-date">
+                        <:label></:label>
+                        <:content>
+                        <AuControlCheckbox
+                            @checked={{this.isCurrentPositionCheck}}
+                            @label="Huidige positie"
+                            @identifier="current-position"
+                            @onChange={{fn (mut this.isCurrentPositionCheck)}}
```
Windvis, line 193: I think we need to create an action for this. If the user selects a date, and then checks this check box, the date is kept. So if they then save, the endDate will still be saved, even though they checked the checkmark. So we either need to clear the endDate when they check the box, or we clear it when saving and the checkbox was checked.

I think the clear on check version is more intuitive since that provides visual feedback to the user.

I know it works like this in the creation page, but we can update that later.

app/router.js
```diff
@@ -28,7 +28,9 @@ Router.map(function () {
         this.route(
           'minister',
           { path: '/bedienaar/:ministerId/' },
-          function () {}
+          function () {
+            this.route('edit');
```
Windvis, line 32: I just noticed that the mandatory edit route is in the administrative unit module instead of the person one ([for example](https://dev.organisaties.abb.lblod.info/bestuurseenheden/a7164f51e0e68820b9d2d4d53aaf48f9/bestuursorganen/44883739b088afd2f1360f8470f30233/mandataris/ded98420a8e608689cb9492ced914342)). So this PR isn't consistent with that. I think it makes sense to move that one to the person module as well, but it requires more changes:

The problem with the current implementation in this PR is that the user always returns to the person position details page, even if they come from the administrative unit. This is weird, especially if the user presses cancel:

1. Go to administrative units minister page
2. Click bewerk
3. Click cancel or save
4. Notice that the user is redirected to the person position details page

If we keep it in the person module, we probably need to implement some redirect logic similar to the person creation page.

app/templates/people/person/positions/minister/edit.hbs
```diff
@@ -0,0 +1,259 @@
...
+                    </:left>
+                </Card.Columns>
+                </:card>
+            </EditCard>
+
+            {{! TODO: add "Voorwaarde" section once file uploads are supported }}
+
+            <AuButtonGroup>
+                <AuButton
+                    @iconAlignment="left"
```
Windvis, line 243: this can be removed since no icon is needed.

Windvis (changes requested):

app/controllers/people/person/positions/minister/edit.js
```diff
@@ -0,0 +1,61 @@
...
+  SELF_FINANCED: '997073905f839ac6bafe92b76050ab0b',
+  FOD_FINANCED: '9d6f49b3d923b437ec3a91e8b5fa6885',
+};
+
+export default class PeoplePersonPositionsMinisterEditController extends Controller {
+  @service router;
+  @tracked willReceiveFinancing;
+  @tracked isCurrentPositionCheck;
+
+  get isCurrentPosition() {
```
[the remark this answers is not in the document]
gcapurso: I copied that logic from the edit mandataris page, I think that it should be changed there too.

app/templates/people/person/positions/minister/edit.hbs
gcapurso: Same as before. I did to make it consistent with the mandataris edit page.

app/router.js
gcapurso: I discussed with anita about that and given the "confusion" related to the path to reach the ministers and the ground inconsistency with reference to mandataris we should to put them here for now.

Windvis: I'm not sure what you mean with 'put them there for now'. But like I said, if we keep it in the person module I would add the same redirect system that we use when creating a person. Unless you get confirmation from Yassin or the Design team that it isn't needed.

gcapurso: Yes the idea is to discuss with Yassin about that after the standup :).

gcapurso: Yessin confirm that we should redirect to the ministers list in the administrative unit.  I should use something like that, right?

```
    if (this.redirectUrl) {
      // When passing a url the query params are ignored so we add the person id manually for now
      this.router.transitionTo(`${this.redirectUrl}?personId=${personId}`);
    } else {
      this.router.transitionTo('people.person', personId);
    }
```

Windvis: Yes, the personId part isn't needed here though.

So the general idea is to add a `redirectUrl` to this edit page, which is reset when leaving the page. Other pages can then set this query parameter when navigating to the edit page.

The edit page itself can then use this query parameter if it is set, or use the person route if it is not.

So it's very similar to the person creation page setup, but without the need to pass an id when redirecting back.

Let me know if you need more information.

gcapurso: I did it but only for the "Bewerk" link in the ministers list page. Otherwise if the user first visit the position details page and than try to modify it through the "Bewerk" link on the top-righ, he'll will be redirected again to the details page where he was before. It makes more sense like that, do you agree?

app/controllers/people/person/positions/minister/edit.js
```diff
@@ -53,9 +68,18 @@ export default class PeoplePersonPositionsMinisterEditController extends Control
     this.model.minister.financing = financing;
     yield this.model.minister.save();
 
-    this.router.transitionTo(
-      'people.person.positions.minister',
-      this.model.minister.id
-    );
+    if (this.redirectUrl) {
+      // When passing a url the query params are ignored so we add the person id manually for now
```
Windvis, line 72: This isn't needed anymore.

app/controllers/people/person/positions/minister/edit.js
```diff
@@ -0,0 +1,85 @@
...
+  setup() {
+    this.willReceiveFinancing =
+      this.model.minister.financing.get('id') === FINANCING_CODE.FOD_FINANCED;
+    this.isCurrentPosition = !this.model.minister.agentEndDate;
+  }
+
+  @action
+  async clearOnCheck() {
+    this.model.minister.agentEndDate = undefined;
+    this.isCurrentPosition = true;
```
Windvis, line 30: The current setup is a bit weird. The checkbox component uses 2 way binding to update the checked state by default, which is why `this.isCurrentPosition` is set to the correct value even if it's being hardcoded to `true` here.

It's confusing when reading the code though, because it looks like it isn't supposed to work.

I would remove the 2 way binding from the component and update the state in this method instead.

app/controllers/people/person/positions/minister/edit.js
```diff
@@ -0,0 +1,85 @@
...
+  queryParams = ['redirectUrl'];
+
+  setup() {
+    this.willReceiveFinancing =
+      this.model.minister.financing.get('id') === FINANCING_CODE.FOD_FINANCED;
+    this.isCurrentPosition = !this.model.minister.agentEndDate;
+  }
+
+  @action
+  async clearOnCheck() {
```
Windvis, line 28: I would rename this to `handleIsCurrentPositionChange` or something since it does more than just clearing the date.

app/templates/people/person/positions/minister/edit.hbs
```diff
@@ -0,0 +1,250 @@
...
+                            @onChange={{fn (mut this.model.minister.agentEndDate)}}
+                            @value={{this.model.minister.agentEndDate}}
+                        />
+                        </:content>
+                    </Item>
+                    <Item @labelFor="end-date">
+                        <:label></:label>
+                        <:content>
+                        <AuControlCheckbox
+                            @checked={{this.isCurrentPosition}}
```
Windvis, line 182: `@checked` is what causes the value to be 2-way-bound and is what actually updates the value at the moment.

I would change it to just a `checked` attribute instead and do the state updates in the action. That makes it more clear how everything works when reading the code.

app/controllers/people/person/positions/minister/edit.js
```diff
@@ -0,0 +1,85 @@
...
+    if (this.redirectUrl) {
+      // When passing a url the query params are ignored so we add the person id manually for now
+      this.router.transitionTo(this.redirectUrl);
+    } else {
+      this.router.transitionTo(
+        'people.person.positions.minister',
+        this.model.minister.id
+      );
+    }
+  }
```
Windvis, line 80: This code and the cancel code are the same, and I think it will always will be the case, so I would extract it into a separate method that can be used in both methods.

gcapurso: Not sure what do you mean with "cancel code".

Windvis: The redirect code in the cancel method.

app/controllers/people/person/positions/minister/edit.js
```diff
@@ -0,0 +1,83 @@
...
+  handleTransitionTo() {
+    if (this.redirectUrl) {
+      this.router.transitionTo(this.redirectUrl);
+    } else {
+      this.router.transitionTo('people.person.positions.minister');
+    }
+  }
+
+  @action
+  async handleIsCurrentPositionChange() {
```
Windvis, line 36: one last thing, didn't notice this before. the `async` isn't needed here.

Windvis (approved):

PR #159, opened 2021-09-08
Some required fields updates

Windvis:
This pr:
 - prevents the users from creating a minister position before selecting a function role first
 - adds required field styling as shown here: https://appuniversum.github.io/ember-appuniversum/docs/patterns/au-card-edit
 - updates the previous implemented required field on the mandatory creation page

Note: This is temporary / quick validation fix. The real validation implementation will include error messages and other data validations.

### DIFF
--- a/app/components/edit-card.hbs
+++ b/app/components/edit-card.hbs
@@ -14,6 +14,12 @@
         )
         to="card"
       }}
+
+      {{#if @containsRequiredFields}}
+        <AuHelpText @skin="secondary" class="au-u-margin-top">
+          * Verplicht veld
+        </AuHelpText>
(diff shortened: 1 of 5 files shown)


Dietr (approved):
Looks good. I do prefer error messages to using a disabled button so you get a clear warning about the required fields. But this is probably out of scope for this PR.

Windvis:
@Dietr Yes indeed, this is the basic / quick "validation" implementation until the actual validations are implemented. Ones those are implemented the button won't be disabled and the user will see error messages when clicking the button and the state is invalid. :+1:

gcapurso (approved):

PR #162, opened 2021-09-09
Accessibility audit, fixes and accessibility statement update

Dietr:
Fixes:
- add alt text to header image
- fix focus on legal pages when navigation
- add "external link" notice to pages opening a new window
- add label id's for KBO & Sharepoint

Update:
- accessibility statement: add information about missing labels for ember-power-select

### DIFF
--- a/app/components/static-page.hbs
+++ b/app/components/static-page.hbs
@@ -6,7 +6,7 @@
     sizes="50vw"
     src="/assets/images/loket-header-1600.jpg"
     srcset="/assets/images/loket-header-320.jpg 320w, /assets/images/loket-header-1024.jpg 1024w, /assets/images/loket-header-1600.jpg 1600w"
-    alt="" role="presentation"
+    alt="een laptop met daarop het vlaanderen icoon"
   />
 </AuContentHeader>
 
--- a/app/templates/contact.hbs
+++ b/app/templates/contact.hbs
@@ -11,7 +11,7 @@
         class="au-c-link"
         target="_blank"
         rel="noopener noreferrer"
-      >handleiding</a>.
+      >handleiding (externe link)</a>.
     </p>
 
     <form class="au-c-form">
--- a/app/templates/legal/accessibilitystatement.hbs
+++ b/app/templates/legal/accessibilitystatement.hbs
@@ -1,6 +1,6 @@
 {{page-title "Toegankelijkheidsverklaring"}}
 
-<AuContent class="au-u-3-4@medium">
+<AuContent class="au-u-3-4@medium" {{auto-focus}} tabindex="-1">
   <AuHeading>Toegankelijkheidsverklaring</AuHeading>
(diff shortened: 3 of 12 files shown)


app/components/static-page.hbs
```diff
@@ -6,7 +6,7 @@
     sizes="50vw"
     src="/assets/images/loket-header-1600.jpg"
     srcset="/assets/images/loket-header-320.jpg 320w, /assets/images/loket-header-1024.jpg 1024w, /assets/images/loket-header-1600.jpg 1600w"
-    alt="" role="presentation"
+    alt="een laptop met daarop het vlaanderen icoon"
```
Windvis, line 9: We removed that alt text by choice, actually. This seems to be a decorative image and I don't think a person using a screenreader needs or wants to know that information?

app/templates/legal/cookiestatement.hbs
```diff
@@ -1,6 +1,6 @@
 {{page-title "Cookieverklaring"}}
 
-<AuContent class="au-u-3-4@medium">
+<AuContent class="au-u-3-4@medium" {{auto-focus}} tabindex="-1">
```
Windvis, line 3: I would explicitly install the auto-focus addon in the app. It only works because ember-appuniversum brings it in at the moment.

app/templates/legal/accessibilitystatement.hbs
```diff
@@ -1,24 +1,24 @@
...
     <li>
-      @TODO
+      <strong>Formulier labels:</strong> de custom select velden hebben geen overeenkomstige labels. <a href="https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html" target="_blank" rel="noopener noreferrer" class="au-c-link">Succescriterium 3.2.2 Labels en instructies (externe link)</a>
     </li>
-  </ul> --}}
+  </ul>
 
   <h2>Opstelling van deze toegankelijkheidsverklaring </h2>
-  <p>Deze toegankelijkheidsverklaring is opgesteld op 26/08/2021.</p>
+  <p>Deze toegankelijkheidsverklaring is opgesteld op 9/09/2021.</p>
```
Windvis, line 21: nitpick, but I would change it to 9/9/2021 or 09/09/2021

app/templates/contact.hbs
```diff
@@ -11,7 +11,7 @@
         class="au-c-link"
         target="_blank"
         rel="noopener noreferrer"
-      >handleiding</a>.
+      >handleiding (externe link)</a>.
```
Windvis, line 14: I've been thinking about creating a dedicated `ExternalLink` or `LinkExternal` component, do you think that is something that can be useful for other projects? In that case it might make sense to add it to ember-appuniversum instead.

app/components/static-page.hbs
Dietr: Ok, removed the alt text. It's a bit strange to use role="presentation" (removes the image completely from the accessibility API and an empty alt text is enough) here but it seems that the linter requires it.

app/templates/contact.hbs
Dietr: Yes, could be helpful as we use it in most projects.

app/components/static-page.hbs
Windvis: Yes, I agree. I discussed this in the ember discord a11y channel and the person responsible for the a11y linting rules ensured me that it was working as intended. After a linking to specific parts of the spec she backtracked and said the rule needed to be updated but I don't think that happened yet.

Windvis (approved):

PR #169, opened 2021-09-20
Implement person validations

Windvis:
This adds basic validations to both the person creation and edit pages. It also includes the groundwork to make adding validations possible.


gcapurso (approved):
Great, now thanks to the error states to the EditCard component we can implement all the logics behind the mandatory fields validation.

PR #171, opened 2021-09-20
Add environment pill

Dietr:
Adds a environment pill in the app-header (solution taken from the kaleidos project).

Additionally: 
- fixes the datepicker overflow bug by updating ember-appuniversum to 0.6.1

### DIFF
--- a/app/styles/components/_environment-pill.scss
+++ b/app/styles/components/_environment-pill.scss
@@ -1,0 +1,37 @@
+/* ==================================
+   #ENVIRONMENT PILL
+   ================================== */
+
+/* Variables
+   ========================================================================== */
+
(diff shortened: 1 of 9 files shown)


app/styles/components/_environment-pill.scss
```diff
@@ -0,0 +1,37 @@
...
+.au-c-environment-pill--local {
+  background-color: $au-environment-local-color;
+  border-color: darken($au-environment-local-color, 7%);
+  color: $au-white;
+}
+
+.au-c-environment-pill--dev {
+  background-color: $au-environment-dev-color;
+  border-color: darken($au-environment-dev-color, 7%);
+  color: white;
```
Windvis, line 30: I guess this should also be `$au-white`?

PR #173, opened 2021-09-20
Use a dedicated component and retrieve the name from the environment config

Windvis:


config/environment.js
```diff
@@ -21,6 +21,7 @@ module.exports = function (environment) {
 
     appName: 'Organisatieportaal',
     contactEmail: 'organisaties.abb@vlaanderen.be',
+    environmentName: '{{ENVIRONMENT_NAME}}', // Supported values: LOCAL, DEV, QA
```
Windvis, line 24: This [will be replaced](https://github.com/mu-semtech/static-file-service#how-to-configure-an-ember-application-at-runtime-using-environment-variables) by the actual value on the server. We just need to set the environment variable on the DEV and QA server.

Dietr (approved):
Would this also work as a component in ember-appuniversum?

Windvis:
Not without some changes. Addons can't import the environment config like that.

Should it behave exactly like this in appuniversum, or should it be more flexible?

Dietr:
Ok, would be nice to have the same component in GN but I guess we can add the same component there. Not really necessary to create it in appuniversum.

PR #180, opened 2021-10-04
Feature/legal structure new

gcapurso:
https://binnenland.atlassian.net/secure/RapidBoard.jspa?rapidView=56&projectKey=OP&modal=detail&selectedIssue=OP-732

Based on the details page design: 
https://www.figma.com/file/P8aIfSN0f1CctxBGYdeyKe/ERE-v1?node-id=964%3A685

and the edit page.

### DIFF
--- a/app/controllers/administrative-units/administrative-unit/legal-structures/new.js
+++ b/app/controllers/administrative-units/administrative-unit/legal-structures/new.js
@@ -19,29 +19,34 @@ export default class AdministrativeUnitsAdministrativeUnitLegalStructuresNewCont
       structuredIdentifier,
     } = this.model;
 
-    address.fullAddress = combineFullAddress(address);
-    yield address.save();
+    yield associatedStructure.validate();
+    yield legalType.validate();
 
-    yield legalType.save();
+    if (associatedStructure.isValid && legalType.isValid) {
+      address.fullAddress = combineFullAddress(address);
+      yield address.save();
 
-    yield structuredIdentifier.save();
-    registration.structuredIdentifier = structuredIdentifier;
-    yield registration.save();
+      yield legalType.save();
 
-    associatedStructure.address = address;
-    associatedStructure.legalType = legalType;
-    associatedStructure.registration = registration;
-    yield associatedStructure.save();
+      yield structuredIdentifier.save();
+      registration.structuredIdentifier = structuredIdentifier;
+      yield registration.save();
 
-    this.model.administrativeUnit.associatedStructures.pushObject(
-      associatedStructure
-    );
-    yield this.model.administrativeUnit.save();
+      associatedStructure.address = address;
+      associatedStructure.legalType = legalType;
+      associatedStructure.registration = registration;
+      yield associatedStructure.save();
 
-    this.router.replaceWith(
-      'administrative-units.administrative-unit.legal-structures.legal-structure',
-      associatedStructure.id
-    );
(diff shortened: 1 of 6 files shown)


app/controllers/administrative-units/administrative-unit/legal-structures/new.js
```diff
@@ -0,0 +1,63 @@
...
+      yield structuredIdentifier.save();
+      registration.structuredIdentifier = structuredIdentifier;
+      yield registration.save();
+
+      associatedStructure.address = address;
+      associatedStructure.legalType = legalType;
+      associatedStructure.registration = registration;
+      yield associatedStructure.save();
+
+      this.model.administrativeUnit.associatedStructures.pushObject(
```
Windvis, line 40: I would destructure `administrativeUnit` in the same way as the other model properties on line 14-20.

app/controllers/administrative-units/administrative-unit/legal-structures/new.js
```diff
@@ -0,0 +1,63 @@
...
+      );
+    }
+  }
+
+  reset() {
+    this.removeUnsavedRecords();
+  }
+
+  removeUnsavedRecords() {
+    this.model.associatedStructure.rollbackAttributes();
```
Windvis, line 57: It seems the new record isn't destroyed like this if it's wrapped in a changeset. I would use `.destroyRecord()` instead. That probably makes sense for all of these records here.

app/templates/administrative-units/administrative-unit/legal-structures/new.hbs
```diff
@@ -0,0 +1,145 @@
+<div class="au-c-body-container au-c-body-container--scroll">
+  <div class="au-o-box au-o-flow au-o-flow--large">
+    <PageHeader>
+      <:title>Voeg juridische structuur toe</:title>
```
Windvis, line 4: I think this should be "Voeg **verbonden** juridische structuur toe"

app/controllers/administrative-units/administrative-unit/legal-structures/new.js
Windvis: You can verify by going to the "new" page, canceling and then checking the data tab in the Ember inspector. It will show unsaved records for the associated-legal-structure and legal-form-type models.

Windvis (changes requested):

app/controllers/administrative-units/administrative-unit/legal-structures/new.js
gcapurso: `.destroyRecord()` gave me problem, I got "Too many recursions" error. Any suggestion?

app/templates/administrative-units/administrative-unit/legal-structures/new.hbs
gcapurso: You are right. I'll fix it in the "edit" page too in the new edit PR that solve the save bug.

app/controllers/administrative-units/administrative-unit/legal-structures/new.js
gcapurso: PS. We used `.rollbackAttributes()` in the person creation page too (in which we include the validation changeset) probably we should change there too.

gcapurso: Debug postoponed.

Windvis (approved):

PR #184, opened 2021-10-05
Feature/sites validations

gcapurso:


Windvis (changes requested):
(I handled these things in the commits I'm about to push)

app/validations/address.js
```diff
@@ -0,0 +1,48 @@
...
+  const REQUIRED_MESSAGE = 'Vul het volledige adres in';
+
+  return {
+    street: validatePresence({
+      presence: true,
+      ignoreBlank: true,
+      message: REQUIRED_MESSAGE,
+      on: isAlwaysRequired
+        ? null
+        : ['number', 'postcode', 'municipality', 'province'],
```
Windvis, line 13: If isAlwaysRequired = `false` then we the field is only required if one of the other address fields are filled in (which is what "on" does.

app/validations/site.js
```diff
@@ -0,0 +1,65 @@
+import {
+  validatePresence,
+  validateFormat,
+} from 'ember-changeset-validations/validators';
+import { validateConditionally } from 'frontend-contact-hub/validators/validate-conditionally';
+
+export const contactValidations = {
+  telephone: validateFormat({
+    allowBlank: true,
+    regex: /^\+32[0-9]{8,9}$/,
```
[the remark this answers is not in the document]
gcapurso: I saw your regex, in that case we are not forcing a given lenght about a phone number, we accept spaces and the prefix is facultative. Is that what we want?

app/validations/site.js
```diff
@@ -0,0 +1,65 @@
...
+    message: 'Enkel een plusteken en cijfers toegelaten.',
+  }),
+  website: validateFormat({
+    allowBlank: true,
+    type: 'url',
```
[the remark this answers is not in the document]
gcapurso: The regex is fine! :+1:

app/validations/address.js
gcapurso: I didn't understand the logic: then in that case if we didn't fill any of the fields, none is required?

Windvis: https://github.com/poteto/ember-changeset-validations#on-option-for-presence

PR #187, opened 2021-10-15
Deep linking support

Windvis:
This implements direct linking to people and "worship administrative units".

It can be tested by using the following url: `http://localhost:4200/redirect?resource=add-resource-uri-here`

If the type of the resource is supported it should go to the correct page, else it will show the index page instead for now.

Requires this branch in the backend app: https://github.com/lblod/app-contact-hub/pull/57

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -41,6 +41,7 @@ module.exports = function (environment) {
     },
 
     showAppVersionHash: process.env.SHOW_APP_VERSION_HASH === 'true',
+    uriInfoServiceUrl: '/uri-info',
 
     userManual: {
       general:
--- a/app/redirect-handlers/administrative-unit.js
+++ b/app/redirect-handlers/administrative-unit.js
@@ -1,0 +1,10 @@
+import { inject as service } from '@ember/service';
+import RedirectHandler from './redirect-handler';
+
+export default class AdministrativeUnitRedirectHandler extends RedirectHandler {
+  @service router;
+
+  redirect(uuid) {
+    this.router.replaceWith('administrative-units.administrative-unit', uuid);
+  }
+}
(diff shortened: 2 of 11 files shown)


app/services/direct-link.js
```diff
@@ -0,0 +1,87 @@
...
+  let triple = uriInfo.triples.find(
+    ({ predicate }) => predicate === RDF_TYPE_URL
+  );
+
+  return triple?.object?.value;
+}
+
+function getUuid(uriInfo) {
+  const MU_UUID_URL = 'http://mu.semte.ch/vocabularies/core/uuid';
+  let triple = uriInfo.triples.find(
```
Windvis, line 66: This assumes there will only be 1 type which isn't always the case. I'm not sure how to handle this though, other than looping through all the results until a redirect-handler is found, or mapping all classes to a redirect handler.

config/environment.js
```diff
@@ -41,6 +41,7 @@ module.exports = function (environment) {
     },
 
     showAppVersionHash: process.env.SHOW_APP_VERSION_HASH === 'true',
+    uriInfoServiceUrl: '/uri-info',
```
Windvis, line 44: Not sure about the name but I didn't want to hardcode this.

app/services/direct-link.js
```diff
@@ -0,0 +1,87 @@
...
+          resourceUri,
+        }
+      );
+    }
+  }
+}
+
+function getRdfType(uriInfo) {
+  const RDF_TYPE_URL = 'http://www.w3.org/1999/02/22-rdf-syntax-ns#type';
+  let triple = uriInfo.triples.find(
```
Windvis, line 57: This assumes there will only be 1 type which isn't always the case. I guess we should loop through all the results until a redirect-handler is found?

Windvis: Added support for multiple rdf types. Now the first one that is mapped to a RedirectHandler class will be used.

Windvis:
I'm merging this myself for now. It seems to work as expected. I just wanted some extra eyes on the implementation but we can always refactor in the future.

PR #200, opened 2021-11-05
Integrate CRAB

nbittich:


app/components/address-register-selector.js
```diff
@@ -0,0 +1,63 @@
...
+
+    if (addressSuggestion) {
+      const addresses = yield this.addressRegister.findAll(addressSuggestion);
+      if (!this.sourceCrab) {
+        this.sourceCrab = yield this.store.findRecord(
+          'concept',
+          'e59c97a9-4e95-4d65-9696-756de47fbc1f'
+        );
+      }
+      // TODO: this should probably be fixed in the API itself (, if possible)
```
nbittich, line 39: it's crab itself i think that has duplicates (uri are differents if i remember, but address seems to be exactly the same)

nbittich (commented):
lgtm

Windvis:
> it's crab itself i think that has duplicates (uri are differents if i remember, but address seems to be exactly the same)

Yea, but the backend service can do the filtering already, right?

> lgtm
Cool, I'll rebase on master and do some squashing

Windvis:
@nbittich I added the co-authored comment (with `nordine@bittich.be` as email) but for some reason it doesn't detect your Github account (it only shows your name). Do you use a different email address by any chance?

nbittich:
> @nbittich I added the co-authored comment (with `nordine@bittich.be` as email) but for some reason it doesn't detect your Github account (it only shows your name). Do you use a different email address by any chance?

i think my email address on github is contact@bittich.be

nbittich:
> @nbittich ah, I see. you might want to add `nordine@bittich.be` to your list of emails in the settings. It seems your commits are signed with that email address so at the moment none of those are linked to your account 😄.

ah didn't know that, I added it now

PR #201, opened 2021-11-09
Feature/yasgui

nbittich:


Windvis (changes requested):
I think you are using npm 7+ which causes a lot of changes to the lock file that aren't really needed. This is also the reason why CI fails since we are still using node 12 + npm 6 there.

I would try to use the same versions on your dev machine to prevent this from happening (take a look at [volta](https://volta.sh/) in case you don't know it yet).

app/templates/sparql.hbs
```diff
@@ -0,0 +1,10 @@
+{{page-title "Sparql"}}
+<div class="au-c-body-container au-c-body-container--scroll">
+  <div class="au-o-region-large">
+    <div class="au-o-layout">
+      <AuContent @skin="large">
```
Windvis, line 5: I don't think this component is needed here?

app/templates/sparql.hbs
```diff
@@ -0,0 +1,10 @@
+{{page-title "Sparql"}}
+<div class="au-c-body-container au-c-body-container--scroll">
```
Windvis, line 2: Optional but might be nice. You could wrap all this in a `StaticPage` component which will add the header and footer like on the other "static pages". Might look nicer?

app/modifiers/yasgui.js
```diff
@@ -0,0 +1,22 @@
+import { modifier } from 'ember-modifier';
+import Yasgui from '@triply/yasgui';
```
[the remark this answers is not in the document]
Windvis: For science I created a production build of this branch and the one I'm working on:

It seems to add about a MB of new JS code:
` - dist/assets/chunk.323.93479c7c76e3ee2084ba.js: 1.02 MB (290.71 KB gzipped) `

It also adds about 50KB of css (9KB compressed):
My branch:
` - dist/assets/frontend-contact-hub-63fd220718901a3fd09be84aee2a2087.css: 106.12 KB (15.98 KB gzipped)`
This branch:
 `- dist/assets/frontend-contact-hub-c75556fdaaa72570d54b33e8ead312e1.css: 147.87 KB (24.27 KB gzipped)`

So it seems to be a rather heavy library. Shipping that to all users should be avoided I think.

app/modifiers/yasgui.js
```diff
@@ -0,0 +1,27 @@
...
+    : `
+        SELECT * WHERE {
+            ?s ?p ?o .
+        } LIMIT 10
+    `;
+
+export default modifier(function yasgui(element /*, params, hash*/) {
+  Promise.all([
+    import('@triply/yasgui'),
+    import('@triply/yasgui/build/yasgui.min.css'),
```
Windvis, line 16: I also tried lazy-loading the css and that seems to work. Saves another 9KB for users who don't visit that page :smile:

Windvis (approved):

PR #215, opened 2021-11-24
Feature/province is mandatory

nbittich:


app/components/address-search.hbs
```diff
@@ -1,65 +1,48 @@
...
-          @options={{this.addressesWithBusNumbers}}
-          @onChange={{this.handleBusNumberChange}}
-          @disabled={{this.isBusNumberSelectionDisabled}}
-          @placeholder={{if
-            this.showBusNumbersNotAvailableMessage
-            "Geen busnummer beschikbaar"
-          }}
-        />
+{{#if this.isAddressSearchMode}}
+<Item @labelFor="{{id}}-address-search" @required={{@required}} @errorMessage={{if @error "Selecteer een adres" }}
```
Windvis, line 3: Could you please disable the auto-formatter in your editor or make sure it uses Prettier :smile:. It seems to be adding a lot of changes that are unrelated to the work.

app/components/address-search.hbs
```diff
@@ -1,65 +1,48 @@
...
-  {{/if}}
+<Item>
+  <:content>
+    <AuButton @skin="link" {{on "click" this.toggleInputMode}}>
+      Vind adres in lijst
+    </AuButton>
+  </:content>
+</Item>
+{{/if}}
+{{yield to="common-input"}}
```
Windvis, line 47: I would move this above the `Vind adres in lijst` link maybe? I think that will look nicer. It should be possible to yield the block from multiple places so that the item appears in both "modes".

app/validations/address.js
```diff
@@ -36,5 +36,13 @@ export function getAddressValidations(isAlwaysRequired = false) {
         ? null
         : ['street', 'number', 'postcode', 'province'],
     }),
+    province: validatePresence({
```
Windvis, line 39: I think this will make it required again for the position pages as well?

app/templates/administrative-units/administrative-unit/governing-bodies/governing-body/mandatory/edit.hbs
```diff
@@ -148,21 +148,17 @@
                         />
                       </:content>
                     </Item>
-                    <Item
-                      @labelFor="contact-address-province"
-                      @errorMessage={{@model.address.error.province.validation}}
-                    >
+                  </:manual-address-input>
+                  <:common-input>
+                    <Item @labelFor="contact-address-province" @required={{true}}
```
Windvis, line 153: I don't think it needed to be required on the "position" pages. So the mandatory and minister creation and edit pages shouldn't make this required, but maybe I misunderstood in the meeting.

Windvis (commented):
I think this should target the master branch since we want to release it as a "hotfix" right?

nbittich:
> I think this should target the master branch since we want to release it as a "hotfix" right?

forgot that yes, fixed

app/validations/address.js
nbittich: waiting for reply of sofie

app/validations/address.js
```diff
@@ -36,5 +36,13 @@ export function getAddressValidations(isAlwaysRequired = false) {
...
+      presence: isAlwaysRequired,
```
Windvis, line 40: This won't work as expected. The wording of this property is confusing. If set to false, this will only be valid if the field doesn't contain any data. But because of the `on` config, that will only trigger if any of the listed fields have data.

So if you enter a street name, and also select a province, the province will show an error.

For now it seems that the only places that require a province also have `isAlwaysRequired` set to true so we can take a shortcut and use that flag for the province as well. (That will fall apart if we ever need an `optional` address that requires a province if some address data is filled, but for now that's not the case).

Adding this province object conditionally based on the `isAlwaysRequired` should do the trick I think.
Pseudo / untested:
```js
let addressValidations = { //all validations but not the province one }

let isProvinceRequired = isAlwaysRequired;
if (isProvinceRequired) {
  addressValidations.province = validatePresence({
    presence: true,
    ignoreBlank: true,
    message: REQUIRED_MESSAGE,
  });
}

return addressValidations;
```

nbittich: thx, i'll test it and let you know

PR #218, opened 2021-11-25
province is required in some cases

nbittich:


app/validations/address.js
```diff
@@ -37,4 +37,15 @@ export function getAddressValidations(isAlwaysRequired = false) {
         : ['street', 'number', 'postcode', 'province'],
     }),
   };
+  if (isProvinceRequired) {
+    addressValidation.province = validatePresence({
+      presence: isAlwaysRequired,
```
Windvis, line 42: I would change that to `true` since it's always true here anyways.

app/validations/address.js
```diff
@@ -37,4 +37,15 @@ export function getAddressValidations(isAlwaysRequired = false) {
...
+      ignoreBlank: true,
+      message: REQUIRED_MESSAGE,
+      on: isAlwaysRequired
```
Windvis, line 45: This isn't needed anymore so I would remove it.

Windvis (changes requested):

Windvis (approved):

PR #225, opened 2021-12-07
Show all related organisations of a central worship service

nbittich:


app/routes/administrative-units/administrative-unit/core-data.js
```diff
@@ -24,7 +23,7 @@ export default class AdministrativeUnitsAdministrativeUnitCoreDataRoute extends
           'primary-site.contacts',
           'is-sub-organization-of',
           'is-associated-with',
-          'sub-organizations',
+          //'sub-organizations',
```
Windvis, line 26: I would remove it if it's not actually needed anymore.

app/routes/administrative-units/administrative-unit/core-data/index.js
```diff
@@ -2,10 +2,16 @@ import Route from '@ember/routing/route';
 
 export default class AdministrativeUnitsAdministrativeUnitCoreDataIndexRoute extends Route {
   model() {
+    const { id } = this.paramsFor('administrative-units.administrative-unit');
     return {
       administrativeUnit: this.modelFor(
         'administrative-units.administrative-unit.core-data'
       ),
+      subOrganizations: this.store.query('organization', {
```
Windvis, line 10: I would await this call. At the moment the page will be rendered before the data is loaded.

An alternative is to use a concurrency task and show a loading spinner while the data is loading. Similar to the organizations setup.

Windvis (changes requested):

Windvis (approved):
Nice!

nbittich:
wait before merging, not sure it works XD

nbittich:
now it should be good, missing return in the task

PR #226, opened 2021-12-09
Implement the change event details page

Windvis:
~~This builds on the changes in #222 so that one needs to be reviewed and merged first.~~ #222 was merged.


app/routes/administrative-units/administrative-unit/change-events/details/index.js
```diff
@@ -0,0 +1,66 @@
...
+
+    let changeEventType = await changeEvent.type;
+
+    return {
+      administrativeUnit,
+      changeEvent,
+      currentChangeEventResult,
+      isMergerChangeEvent: changeEventType.id === CHANGE_EVENT_TYPE.MERGER,
+      shouldShowExtraInformationCard:
+        shouldShowExtraInformationCard(changeEventType),
```
Windvis, line 45: It's probably better to move these to the controller since they are derived state.

app/templates/administrative-units/administrative-unit/change-events/details/index.hbs
```diff
@@ -0,0 +1,134 @@
...
+    </DataCard>
+
+    {{#if @model.shouldShowExtraInformationCard}}
+      <DataCard>
+        <:title>{{@model.changeEvent.type.label}}</:title>
+        <:card as |Card|>
+          <Card.Columns>
+            <:left as |Item|>
+              {{#each
+                @model.changeEvent.originalOrganization
```
Windvis, line 90: This will show the organizations in a sorted way, but the design shows the current organization on top. Not sure if that is a requirement though. That means we need a custom sort helper or do the sorting in the model hook.

app/templates/administrative-units/administrative-unit/change-events/details/index.hbs
```diff
@@ -0,0 +1,134 @@
...
+              {{/each}}
+            </:left>
+            <:right as |Item|>
+              {{#if @model.isMergerChangeEvent}}
+                <Item>
+                  <:label>Resulterende organisatie</:label>
+                  <:content>
+                    <AuLink
+                      @route="administrative-units.administrative-unit"
+                      @model={{@model.changeEvent.resultingOrganization.firstObject.id}}
```
Windvis, line 113: In the data model there can be multiple "resulting organizations" but in our application there can only ever be 1.

app/templates/administrative-units/administrative-unit/change-events/details/index.hbs
Windvis: I'll keep it as-is for now and ask the UX team when they are back from holidays.

PR #228, opened 2021-12-09
Hide edit/add buttons for lezer

nbittich:


Windvis (changes requested):
I'm honestly not a big fan of the current implementation. It works, but it doesn't really scale when the application becomes more complex and different users start using the app.

I think https://github.com/Bagaar/ember-permissions would be a nicer way to set this up. It was made for this use case and it's easier to set up more fine grained permissions in the future. I didn't know this was planned this sprint or else I would have documented it in the issue.

The roles can be mapped to a list of permissions (for now only `edit` I guess) and the helpers can be used instead of the `<SecuredArea>` component.

The repeated code in the before model hooks would be replaced by setting the route permissions and handling the `route-access-denied` event.

Now, I assume everything works so it might not be worth redoing everything since we can still refactor it in the future. It would have been nice if you pinged me before starting the implementing though.

nbittich:
> I'm honestly not a big fan of the current implementation. It works, but it doesn't really scale when the application becomes more complex and different users start using the app.
> 
> I think https://github.com/Bagaar/ember-permissions would be a nicer way to set this up. It was made for this use case and it's easier to set up more fine grained permissions in the future. I didn't know this was planned this sprint or else I would have documented it in the issue.
> 
> The roles can be mapped to a list of permissions (for now only `edit` I guess) and the helpers can be used instead of the `<SecuredArea>` component.
> 
> The repeated code in the before model hooks would be replaced by setting the route permissions and handling the `route-access-denied` event.
> 
> Now, I assume everything works so it might not be worth redoing everything since we can still refactor it in the future. It would have been nice if you pinged me before starting the implementation though.

Well, I set 8 story points for the ticket and did implement it in 1 hour lol. I knew it will take more time than that after the PR would be open. to me it was easier to implement it the way I would, open the discussion through PR and then refactor accordingly. I know you are busy with the change events so didn't want to interrupt. I still have time to try out ember permission tbf.

app/routes/administrative-units/administrative-unit/core-data/edit.js
```diff
@@ -10,6 +10,15 @@ import worshipAdministrativeUnitValidations, {
 
 export default class AdministrativeUnitsAdministrativeUnitCoreDataEditRoute extends Route {
   @service store;
+  @service currentSession;
+  @service session;
```
Windvis, line 14: The session service isn't used so I wouldn't inject it.

app/routes/administrative-units/administrative-unit/core-data/edit.js
```diff
@@ -10,6 +10,15 @@ import worshipAdministrativeUnitValidations, {
...
+  @service router;
+
+  beforeModel() {
+    if (!this.currentSession.canEdit) {
+      this.router.transitionTo('route-not-found', { wildcard: 'not-found' });
```
Windvis, line 19: I think we went with dutch wording for the path names. So I would go for something like this instead:
```suggestion
      this.router.transitionTo('route-not-found', { wildcard: 'pagina-niet-gevonden' });
```

Windvis (changes requested):
Some final minor things!

Windvis (approved):
Also did a quick test locally and everything seems to work as expected. Nice!

PR #229, opened 2021-12-14
Allow users to easily report wrong data

nbittich:
see : https://abb-vlaanderen.gitbook.io/doc-organisatieportaal/overview/editing-information-and-card-views/feature-passport-report-wrong-data

### DIFF
--- a/app/templates/administrative-units/administrative-unit/local-involvements/edit.hbs
+++ b/app/templates/administrative-units/administrative-unit/local-involvements/edit.hbs
@@ -4,8 +4,8 @@
     <:title>Bewerk betrokken lokale besturen</:title>
     <:subtitle>{{@model.administrativeUnit.name}}</:subtitle>
     <:action>
-      <div class="au-o-flow au-d-component">
-        <div>
+      <div>
+        <AuButtonGroup class="au-c-button-group--align-right">
           <AuButton
             @loading={{this.save.isRunning}}
             @disabled={{this.save.isRunning}}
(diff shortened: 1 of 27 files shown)


app/templates/administrative-units/administrative-unit/local-involvements/edit.hbs
```diff
@@ -4,20 +4,27 @@
...
-        Opslaan
-      </AuButton>
-      <AuLink
-        @route="administrative-units.administrative-unit.local-involvements"
-        @skin="button-secondary"
-      >
-        Annuleer
-      </AuLink>
+      <div class="au-o-flow au-d-component">
+        <div>
```
Windvis, line 8: Unrelated to this PR, but could you convert this to the `<AuButtonGroup>` component? That will add proper margin between buttons (at the moment the whitespace is only there because the template includes a space between the components).

I also noticed that the buttons aren't aligned to the right anymore:
image

Sadly the `<AuButtonGroup>` component [doesn't provide an argument for that yet](https://github.com/appuniversum/ember-appuniversum/issues/162) so we will need to add some custom css for now.

Maybe adding something like this to the _shame.scss file is good enough for now? 
```css
// Allow buttons to be aligned on the right side
.au-c-button-group--align-right {
  @include mq($from:medium) {
    justify-content: flex-end;
  }
}
```

Which can then simply be used like this:
```hbs
{{! TODO: Remove the custom class once the following issue is resolved: https://github.com/appuniversum/ember-appuniversum/issues/162 }}
<AuButtonGroup class="au-c-button-group--align-right">...
```

It's not perfect on all screen sizes but I think it's a good starting point.

Windvis (changes requested):

Windvis (approved):

PR #231, opened 2021-12-15
Move the form buttons to the top of the page

nbittich:


app/templates/administrative-units/administrative-unit/ministers/new.hbs
```diff
@@ -5,6 +5,26 @@
...
+          @route="administrative-units.administrative-unit.ministers"
+          @skin="button-secondary"
+        >
+          Annuleer
+        </AuLink>
+      </AuButtonGroup>
+    </:action>
   </PageHeader>
 
   {{#if this.isSelectingTargetPerson}}
```
Windvis, line 30: This check also needs to be added before showing the buttons. At the moment they are visible before a person is selected. The same check is needed on the mandatory creation page.

app/templates/administrative-units/administrative-unit/core-data/edit.hbs
```diff
@@ -4,7 +4,25 @@
       <:title>Bewerk kerngegevens</:title>
       <:subtitle>{{@model.administrativeUnit.name}}</:subtitle>
       <:action>
-        <ReportWrongData />
+        <div>
```
Windvis, line 7: This looks fine but because the buttons aren't nested in the form anymore they aren't linked and clicking the button won't submit the form.

Adding an id to the form and using the [form attribute on the button](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-form) should work and it's probably needed on all the other pages as well.

Windvis (changes requested):

app/templates/administrative-units/administrative-unit/ministers/new.hbs
```diff
@@ -5,6 +5,30 @@
   <PageHeader class="au-u-margin au-u-margin-bottom-large">
     <:title>Voeg bedienaar toe</:title>
     <:subtitle>{{@model.administrativeUnit.name}}</:subtitle>
+
+    <:action>
+      {{#if this.isSelectingTargetPerson}}
```
Windvis, line 10: This check needs to be reversed I think? At the moment the buttons are still visible in the "person selection" stage but not in the form.

Windvis (changes requested):

app/templates/administrative-units/administrative-unit/ministers/new.hbs
nbittich: ah yes sorry, stupid mistake. thought it was an easy one so went too fast, next time i'll be more careful, sorry for  that

nbittich: when I committed  it changed from "if (not" to "unless"

Windvis: Yea, that's because of the [no-negated-condition](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-negated-condition.md) linting rule which has an autofixer (which runs precommit because of husky).

I'm not a big fan of that rule, I think `if not something` is way more readable than `unless something`. But it's part of the default ruleset and we try to not overrule those as much as possible. I hope this default rule changes in the future though and there are more people who have a similar mindset: https://github.com/emberjs/rfcs/issues/714

Windvis (approved):

PR #234, opened 2022-01-05
Implement the change event creation page

Windvis:
requires this backend branch: https://github.com/lblod/app-contact-hub/pull/75


nbittich (approved):
I tested it locally, was able to create a new change event; looks good enough to be merged so that the business can test it and gives feedback

PR #235, opened 2022-01-05
Add a `requireEditPermissions` method to the currentSession service

Windvis:
This method can be used to automatically redirect users to the `route-not-found` page if they don't have the needed permissions. It works very similar to ember-simple-auth's `requireAuthentication` method. 

This version has the benefit that the url doesn't change to "/pagina-niet-gevonden" so it works similar to traditional websites.

Usage:
```diff
  beforeModel() {
-    if (!this.currentSession.canEdit) {
-      this.router.transitionTo('route-not-found', {
-        wildcard: 'pagina-niet-gevonden',
-      });
-    }
+    this.currentSession.requireEditPermissions(transition);
  }
```

This can be rolled out to all the pages that need it if we want.

### DIFF
--- a/app/services/current-session.js
+++ b/app/services/current-session.js
@@ -10,6 +10,7 @@ const EDITOR_ROLES = [
 const READER_ROLES = ['ABBOrganisatiePortaalGebruiker-lezer'];
 
 export default class CurrentSessionService extends Service {
+  @service router;
   @service session;
   @service store;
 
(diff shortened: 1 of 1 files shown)


Windvis:
I extracted this from the change events branch since it's a standalone feature.

Windvis:
This looks interesting: https://github.com/simplabs/ember-error-route

PR #241, opened 2022-01-13
Always set original and resulting organization when creating change events

Windvis:
The data model expects these properties to be set most of the time (with some exceptions), although it wasn't needed for the frontend to work properly.


Windvis:
I converted it back to draft because Yassin needs to confirm if the current logic is ok.

nbittich:
> I converted it back to draft because Yassin needs to confirm if the current logic is ok.
Is it ready to be reviewed?

Windvis:
@nbittich yes, sorry I was going to ping you after the standup but I forgot :smile:.

nbittich (approved):

PR #242, opened 2022-01-13
Only save a decision if it contains data

Windvis:
Saving an empty record will create "empty" triples in the database which isn't a good idea from a data quality perspective.


Windvis:
Draft PR since more testing is needed, but this has a low priority.

Windvis:
Did some more testing and everything seems to work. This also fixes an issue were pressing save didn't work for the `erkenning aangevraagd` change event type since it never has a decision.

nbittich (approved):

PR #244, opened 2022-01-14
Update to the latest madnificent/ember image

Windvis:
This image uses node 16 and npm 8.3 so we pin those versions for volta users and also update the lock file to match that setup.


Windvis:
We still need to update the drone file. It uses the danlynn 3.22 images which uses node 12 / npm 6.

PR #248, opened 2022-01-21
Improve the original organizations selection logic on the change events creation page

Windvis:
Selecting more than 1 original organization is no longer a requirement for the "gebiedswijziging" change event.

### DIFF
--- a/app/routes/administrative-units/administrative-unit/change-events/new.js
+++ b/app/routes/administrative-units/administrative-unit/change-events/new.js
@@ -80,8 +80,23 @@ class FormState {
   set changeEventType(value) {
     this._changeEventType = value;
 
+    if (
+      this.shouldSelectMultipleOriginalOrganizations &&
+      !this.isAddingOriginalOrganization &&
+      this.originalOrganizations.length === 0
+    ) {
+      this.isAddingOriginalOrganization = true;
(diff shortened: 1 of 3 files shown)


app/routes/administrative-units/administrative-unit/change-events/new.js
```diff
@@ -80,12 +80,27 @@ class FormState {
...
     if (this.error?.changeEventType) {
       delete this.error.changeEventType;
       this.error = {
         ...this.error,
       };
     }
+
+    if (this.error?.originalOrganizations) {
+      delete this.error.originalOrganizations;
+      this.error = {
```
nbittich, line 100: what does this?

nbittich (approved):

app/routes/administrative-units/administrative-unit/change-events/new.js
Windvis: this.error is a tracked reference so assigning something to it lets Ember know that the views that use it should be re-rendered. Since deleting a key from the object doesn't trigger that system (without using something like https://github.com/tracked-tools/tracked-built-ins and even then I'm not sure) I treat it as immutable and assign it a new object so that ember knows it needs to re-render.

PR #249, opened 2022-01-26
Fix a problem were saving doesn't work on the local involvements page

Windvis:
We now check if there are any financial organizations before triggering the total percentage validation. This fixes a problem were a valid form was marked invalid and prevented users from saving.


Windvis:
FYI Don't squash and merge this PR please, the separate commits are useful for the history in this case.

nbittich (approved):

PR #250, opened 2022-02-01
Add support for primary and secondary contact information

Windvis:
This PR adds support for adding a secondary telephone number on all pages with contact information or updates the labels of the ones that already supported this.

It expects that the migrations from https://github.com/lblod/app-contact-hub/pull/88 ran already. Once we merge this we can also merge that PR.


Windvis:
There are a lot of commits since I created a commit for each page. I might reduce this by combining all related pages into one commit.

PR #254, opened 2022-02-03
Feature/mu search page person

nbittich:


app/controllers/people/index.js
```diff
@@ -34,14 +36,26 @@ export default class PeopleIndexController extends Controller {
 
   get hasNoResults() {
     return (
-      this.model.loadPeopleTaskInstance.isFinished && this.people.length === 0
+      this.model?.loadPeopleTaskInstance?.isFinished &&
```
Windvis, line 39: Why is the optional chaining needed here? Normally the controller always has the model set before the page is rendered.

app/controllers/people/index.js
```diff
@@ -34,14 +36,26 @@ export default class PeopleIndexController extends Controller {
...
+      this?.people?.length === 0
     );
   }
 
   get hasErrored() {
     return this.model.loadPeopleTaskInstance.isError;
   }
 
+  @action
+  setPosition(event) {
```
Windvis, line 49: nitpick: the argument is the record, not an event. I would rename it to `position` since it is a bit confusing at the moment.

app/controllers/people/index.js
```diff
@@ -34,14 +36,26 @@ export default class PeopleIndexController extends Controller {
...
+    this.position = event?.id;
+  }
+
+  @action
+  setOrganization(event) {
```
Windvis, line 54: same here

app/templates/people/index.hbs
```diff
@@ -4,29 +4,30 @@
...
-          </p>
           <p class="au-o-grid__item">
             <AuLabel for="filter-organization">Organisatie</AuLabel>
-            <AuInput
-              @width="block"
-              @value={{this.organization}}
-              id="filter-organization"
+            <AdministrativeUnitSelect
+              @allowClear={{true}}
+              @selected={{this.org}}
```
Windvis, line 18: This won't work after the page is refreshed. The record is only stored in memory so after a page refresh the record is gone. The id that's stored in a query param needs to somehow be used to retrieve the correct record again.

Some other select components receive the id itself and do the lookup in the list of options, but I don't think that will work here since we use `search` to find results which means simple getters can't be used. I _think_ this might be a use case for [resources](https://github.com/NullVoxPopuli/ember-resources) but I haven't tried that addon myself.

app/templates/people/index.hbs
```diff
@@ -4,29 +4,30 @@
       <div class="au-c-sidebar__content au-o-box">
         <form {{on "submit" this.search}} class="au-o-grid au-o-grid--small">
           <p class="au-o-grid__item au-u-1-2@small au-u-1-1@medium">
```
Windvis, line 6: I think the au-u-* classes can be removed. It looks a bit strange at the moment:

image

Windvis (changes requested):
Nice! The speed increase is _really_ nice. Users will love it.

app/components/position-select.js
```diff
@@ -0,0 +1,34 @@
...
+    if (this.loadPositionTask.isRunning) {
+      return null;
+    }
+
+    let position = this.loadPositionTask.last.value;
+    return position.find((p) => p.id === id);
+  }
+
+  @task *loadPositionTask() {
+    return yield this.store.findAll('board-position');
```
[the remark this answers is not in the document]
nbittich: yeah, I think for now it should stay as is but I'm not really sure. It's what I was discussing in the previous standup, related to OP-483. If we list the minister functies, it won't be reflected to the list page  if someone filters on it anyway right?

app/routes/people/index.js
```diff
@@ -22,34 +23,34 @@ export default class PeopleIndexRoute extends Route {
...
-        size: params.size,
-      },
-      sort: params.sort,
-    };
-
-    if (params.givenName) {
-      query['filter[given-name]'] = params.givenName;
+    const filter = {};
+    if (params.name) {
+      filter['given_name,family_name'] = params.name;
```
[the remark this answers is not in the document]
nbittich: Good point, I added fuzziness

app/templates/people/index.hbs
nbittich: I followed a different approach, but I'm not very comfortable with doing things in the constructor. what do you think? It seems working visually

Windvis: Hmm, I think this will break the other places where we use this component.

Resources are indeed not needed if we assume that the fetching of a record only needs to happen once, on initial render, which is probably the case for us anyways. So we can indeed use the constructor.

I would make some changes though:
- in the constructor we could add a check to see if it is a string (id) and if it is, fetch the record
- rename the `selected` to `loadedRecord` or something so it's more clear what it is and not cause confusion with the selected argument.
- add a getter that does the same check [(similar like some other selects)](https://github.com/lblod/frontend-contact-hub/blob/533e58f537473e708c101061b1e195a1afe209c1/app/components/recognized-worship-type-select.js#L14-L20)
  - if it's a string we check if the task has already finished and either return the record or null
  - if it's an object we return that instead
  - once the user selects an option we can in theory also empty the loaded record from the component state
  
In the controller we could then store the actual selected record in a separate tracked property from the `setOrganization` method and pass it into the component with `or`

`@selected={{or this.selectedOrganization this.organization}}`

I _think_ that should work and the other places we use this should keep working as well.

nbittich: seems working  👍🏼

Windvis (approved):

PR #256, opened 2022-02-07
Show minister position information on the positions page of a person

nbittich:


app/routes/people/person/positions/index.js
```diff
@@ -13,15 +13,48 @@ export default class PeoplePersonPositionsIndexRoute extends Route {
...
+        id: minister.id,
+        startDate: minister.agentStartDate,
+        endDate: minister.agentEndDate,
+        administrativeUnit,
+      });
+    }
     return {
       person,
       positions,
-      ministerPositions: await person.agentsInPosition,
```
nbittich, line 24: I'm not sure if this will break other things @Windvis  ?

Windvis: Probably not, it doesn't seem like it was used anywhere?

Windvis (approved):

PR #259, opened 2022-02-09
Allow users to edit contact information on the person edit page

nbittich:
as discussed with Saska: "The org and position is not editable on there but the contact info should be"

### DIFF
--- a/app/controllers/people/person/personal-information/edit.js
+++ b/app/controllers/people/person/personal-information/edit.js
@@ -1,6 +1,7 @@
 import Controller from '@ember/controller';
 import { dropTask } from 'ember-concurrency';
 import { inject as service } from '@ember/service';
+import { combineFullAddress } from 'frontend-contact-hub/models/address';
 
 export default class PeoplePersonPersonalInformationEditController extends Controller {
   @service router;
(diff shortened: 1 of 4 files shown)


app/controllers/people/person/personal-information/edit.js
```diff
@@ -1,23 +1,45 @@
...
+        }
+        if (secondaryContact.isDirty) {
+          yield secondaryContact.save();
+        }
+      }
+    }
 
-    if (this.model.person.isValid) {
-      yield this.model.person.save();
+    if (person.isValid) {
```
Windvis: I would wrap this check around the contacts loop? Otherwise contacts might be saved while the person itself is invalid.

app/routes/people/person/personal-information/edit.js
```diff
@@ -14,11 +16,31 @@ export default class PeoplePersonPersonalInformationEditRoute extends Route {
...
+    const contacts = [];
+    for (let position of positions) {
+      let address = await position.primaryContact.contactAddress;
+      let contact = {
+        title: 'Contactgegevens: ' + position.title,
+        primaryContact: createValidatedChangeset(
+          position.primaryContact,
+          contactValidations
+        ),
+        address: createValidatedChangeset(address, getAddressValidations),
```
Windvis, line 32:
```suggestion
        address: createValidatedChangeset(address, getAddressValidations()),
```

app/controllers/people/person/personal-information/edit.js
```diff
@@ -1,23 +1,49 @@
...
+        }
+      } else {
+        validContacts = false;
+        break;
+      }
+    }
 
-    if (this.model.person.isValid) {
-      yield this.model.person.save();
+    if (validContacts && person.isValid) {
```
Windvis, line 43: I would change the flow a bit. It's still possible that some records are being saved while not everything is valid.

I would first loop through all the contacts to see if there are any invalid records. If there aren't any and the person is also valid the saving logic can run.

Windvis (changes requested):

app/templates/people/person/personal-information/edit.hbs
```diff
@@ -94,6 +94,21 @@
             </Card.Columns>
           </:card>
         </EditCard>
+        {{#each @model.contacts as |contact|}}
+          <Position::ContactEditCard
+           @title={{contact.title}}
```
Windvis, line 99:
```suggestion
```

Windvis (changes requested):
1 small thing, but other than that it seems good to go.

nbittich:
> 1 small thing, but other than that it seems good to go.

oops, thanks for notifying! totally forgot that one

Windvis (approved):

PR #261, opened 2022-02-10
perf bestuur

nbittich:


app/routes/administrative-units/index.js
```diff
@@ -15,15 +15,16 @@ const CLASSIFICATION = {
 
 export default class AdministrativeUnitsIndexRoute extends Route {
   @service store;
```
Windvis, line 17: This isn't used anymore I think?
```suggestion
```

Windvis (approved):
Haven't tested it but the code looks good to me.

claire-lovisa (approved):
Tested and works as advertised ! ;)

PR #266, opened 2022-02-28
Fix Edit page when primary contact is null

nbittich:
For some position, when you try to edit them, you get an error (e.g => https://dev.organisaties.abb.lblod.info/personen/19a4d659d73809129467ac1c39e7fd02/contactgegevens )
This PR is an attempt to fix it. As we mix two kinds of resources (mandates, ministers), I had to work around a little bit.

### DIFF
--- a/app/controllers/people/person/personal-information/edit.js
+++ b/app/controllers/people/person/personal-information/edit.js
@@ -29,8 +29,7 @@ export default class PeoplePersonPersonalInformationEditController extends Contr
 
     if (valid) {
       for (let contact of contacts) {
-        let { primaryContact, secondaryContact, address, _ember_data } =
-          contact;
+        let { primaryContact, secondaryContact, address, _position } = contact;
         if (address.isDirty) {
           address.fullAddress = combineFullAddress(address);
           yield address.save();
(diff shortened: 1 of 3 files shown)


app/controllers/people/person/personal-information/edit.js
```diff
@@ -29,7 +29,7 @@ export default class PeoplePersonPersonalInformationEditController extends Contr
 
     if (valid) {
       for (let contact of contacts) {
-        let { primaryContact, secondaryContact, address } = contact;
+        let { primaryContact, secondaryContact, address, _position } = contact;
```
claire-lovisa, line 32: I am not sure why we should prefix `_position` here ? I tried to look for some doc about prefixing with an underscore but I didn't find :sweat_smile:

app/routes/people/person/personal-information/edit.js
```diff
@@ -22,16 +25,31 @@ export default class PeoplePersonPersonalInformationEditRoute extends Route {
     );
     const contacts = [];
     for (let position of positions) {
-      let address = await position.primaryContact.contactAddress;
+      let { primaryContact, secondaryContact, _position } = position;
+      if (!primaryContact) {
+        primaryContact = createPrimaryContact(this.store);
+        primaryContact.contactAddress = this.store.createRecord('address');
```
claire-lovisa, line 31: It's a detail, but this line is a little bit redundant as we already create an address line 40 if we don't have one already ?

claire-lovisa (approved):

app/controllers/people/person/personal-information/edit.js
nbittich: I ddidn't have a really good reason 😄 .
I wanted  to say "this is the underlying ember data position that can be either a minister or mandatory" 
Do you think it makes sense?

app/routes/people/person/personal-information/edit.js
nbittich: true 👍🏼

claire-lovisa (approved):
Tested and approved ! :sun_with_face:

PR #268, opened 2022-03-07
fix position details not shown

nbittich:
this fixes OP-1197


claire-lovisa (approved):
Works as advertised :)

PR #269, opened 2022-03-08
Remove werkingsgebieden

claire-lovisa:
Goes together with https://github.com/lblod/app-contact-hub/pull/114


nbittich (approved):
tested locally, works as expected 👍🏼

PR #270, opened 2022-03-09
split first name / last name in filter search person

nbittich:
Bugfix for OP-1196


claire-lovisa (approved):
It's indeed nicer to use like that ! :)

PR #275, opened 2022-03-16
sensitive data

nbittich:
Should be reviewed after:
https://github.com/lblod/privacy-centric-service/pull/3
https://github.com/lblod/app-contact-hub/pull/118


nbittich:
I fixed the first point, but for the second point I've kept it as-is. The reason is we override the 4 fields every time  we update sensitive data, so showing empty input will delete inputs filled that wasn't requested.

As it is no't really clear in FP, I'd prefer to wait for Sofie to have her opinion about it; if we need to change something, it's probably another ticket and FP will have to be updated I think.

I also managed to fix the refresh problem and made it as it was asked in the fp, which wasn't the case before:

```
The data will be available for the user to see until the user refreshes the page. This is how it currently works.
```
so now data of every person requested are available as long as user doesn't refresh page

claire-lovisa (approved):
Really like this change to get rid of the query param ! Smart !

PR #277, opened 2022-03-17
searching in field “naam” with multiple words gives no result

nbittich:
bug fix for OP-1222


claire-lovisa (changes requested):
You might want to keep the `.toLowerCase()`, I tested with `Anglicaanse Kerk English Church van Oostende` and got the following when searching : 
- "Anglicaanse Kerk English Church van Oostende" -> result
- "oostende" : results
- "Oostende" : no result

claire-lovisa (approved):

PR #280, opened 2022-03-21
Rename the project from contact-hub to organization-portal

claire-lovisa:


claire-lovisa:
@nbittich Very good remark, i'll add a task for it so that we don't forget !

PR #285, opened 2022-03-22
bug fix for relation worship admin unit

nbittich:
[OP-1236](https://binnenland.atlassian.net/browse/OP-1236)


claire-lovisa (commented):
It works, but could you remove the relationship from `worship-service.js` as it's already inherited from the worship administrative unit ? Thanks !

claire-lovisa (approved):

PR #287, opened 2022-03-23
change the filter search for name field

nbittich:
tested with both Sofie's and Zelim's comment => OP-1222


claire-lovisa (approved):
Nice find !

PR #295, opened 2022-03-25
typo fixes for sensitive info

nbittich:
Bugfixes for OP-1250 & OP-1251


claire-lovisa (approved):
Not tested but it's only text changes so looks good :)

PR #305, opened 2022-04-01
autosuggestion person

nbittich:
OP-1265


claire-lovisa (changes requested):
When you select a string as first name, like "ra", the last name gets filled with an empty string. If then you add a string to last name, it clears the first name and replaces it by an empty string. I guess we should be able to enter both a string for first and last name ?
And when you clear one field, the second one gets cleaned as well, we might want to only clear one ?

claire-lovisa (approved):

PR #306, opened 2022-04-04
Remove duplicates in person search result

nbittich:
OP-1265 => fix Sofie's comment

### DIFF
--- a/app/components/person-search-by-name.js
+++ b/app/components/person-search-by-name.js
@@ -26,20 +26,24 @@ export default class PersonSearchByNameComponent extends Component {
       dataMapping: (data) => {
         const entry = data.attributes;
         return {
-          given_name: entry.given_name,
-          family_name: entry.family_name,
+          given_name: entry.given_name.toLowerCase(),
+          family_name: entry.family_name.toLowerCase(),
         };
       },
     });
 
-    result = new Set(result.toArray());
     if (searchParams.trim() !== '' && result) {
       let param_object = {
         family_name: '',
         given_name: '',
       };
       param_object[fieldName] = searchParams.trim();
-      return [...[param_object], ...result];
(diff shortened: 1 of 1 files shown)


app/components/person-search-by-name.js
```diff
@@ -26,20 +26,28 @@ export default class PersonSearchByNameComponent extends Component {
...
+      );
     if (searchParams.trim() !== '' && result) {
       let param_object = {
         family_name: '',
         given_name: '',
       };
       param_object[fieldName] = searchParams.trim();
-      return [...[param_object], ...result];
+      const arr = [...[param_object], ...result];
+      console.log(arr);
```
claire-lovisa, line 49: This one should probably be removed ? :p

app/components/person-search-by-name.js
```diff
@@ -26,20 +26,28 @@ export default class PersonSearchByNameComponent extends Component {
...
-
-    result = new Set(result.toArray());
+    result = result
+      .toArray()
+      .filter(
+        (v, i, a) =>
+          a.findIndex((v2) =>
+            ['family_name', 'given_name'].every((k) => v2[k] === v[k])
+          ) === i
+      );
```
claire-lovisa, line 41: Can you maybe add a comment about what this filter does ? It's not straightforward when reading it :)

claire-lovisa (approved):
Little comments, but the PR works as it should ! :)

app/components/person-search-by-name.js
nbittich: indeed! I extracted the code in a separate function and added a documentation to it

app/components/person-search-by-name.js
nbittich: oops 😱

PR #317, opened 2022-04-13
copy mechanism poc

nbittich:
Goal of this POC is to deploy a working example on dev. I tested it locally but I'd really appreciate a second look before deploying it; design is different than on figma, this is normal (will be finalize after demo it to Saska & Sofie)

### DIFF
--- a/app/styles/_shame.scss
+++ b/app/styles/_shame.scss
@@ -135,3 +135,6 @@ table.dataTable tbody th {
 .items-center {
   align-items: center !important;
 }
+.zindex9998 {
+  z-index: 9999;
+}
--- a/app/styles/_shame.scss
+++ b/app/styles/_shame.scss
@@ -135,6 +135,3 @@ table.dataTable tbody th {
 .items-center {
   align-items: center !important;
 }
-.zindex9998 {
-  z-index: 9999;
-}
--- a/app/templates/administrative-units/administrative-unit/ministers/new.hbs
+++ b/app/templates/administrative-units/administrative-unit/ministers/new.hbs
@@ -97,11 +97,6 @@
         </:card>
       </EditCard>
 
-      {{!-- <Position::ContactEditCard
-        @address={{@model.address}}
-        @primaryContact={{@model.contact}}
-        @secondaryContact={{@model.secondaryContact}}
-      /> --}}
       <AuContent>
         <ContactDetails
           @title="Contactgegevens"
--- a/app/templates/administrative-units/administrative-unit/ministers/new.hbs
+++ b/app/templates/administrative-units/administrative-unit/ministers/new.hbs
@@ -33,7 +33,8 @@
 
   {{#if this.isSelectingTargetPerson}}
     <PersonSearch
-      @onSelect={{fn (mut this.targetPerson)}}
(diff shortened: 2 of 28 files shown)


app/templates/administrative-units/administrative-unit/ministers/new.hbs
```diff
@@ -96,12 +97,19 @@
         </:card>
       </EditCard>
 
-      <Position::ContactEditCard
+      {{!-- <Position::ContactEditCard
```
claire-lovisa, line 100: Is the comment still needed for later ?

app/styles/_shame.scss
```diff
@@ -135,3 +135,6 @@ table.dataTable tbody th {
 .items-center {
   align-items: center !important;
 }
+.zindex9998 {
```
claire-lovisa, line 138: I'm not seeing it used in the PR, is it needed ?

claire-lovisa (commented):
It's looking very good !! I wrote down some remarks while testing it, some might not be valid, but here's my thoughts : 
- When I go to a persons's position, select an address and save, I'm redirected to a bestuursorgaan without any mandatarissen / bestuursleden (but I should, because I just updated a position ?)
- The addresses showing in the list, for a person, are the active addresses. Technically, I totally get why it's the case, but jut to be sure, did we expect that older addresses added for that same person would show up ?
- Doing the following when there is already a contact detail in the list will replace it by blank contact details : "Nieuw" -> Don't fill anything -> "Opslaan" (+ it's not persisted, when I refresh the old details are back)
- Suggestion - Reuse the buttons formatting of "Bewerk kerngegevens" to make it stand out a little more ?

nbittich:
Thanks :)

> * When I go to a persons's position, select an address and save, I'm redirected to a bestuursorgaan without any mandatarissen / bestuursleden (but I should, because I just updated a position ?)

It seems to be the same behavior on dev / acc. I don't think I changed anything there 🤔  The address is persisted tho, if you go back to the position, it is correctly selected.


>     * The addresses showing in the list, for a person, are the active addresses. Technically, I totally get why it's the case, but jut to be sure, did we expect that older addresses added for that same person would show up ?

Yes this is required in the FP. only active position's addresses should be listed.

>     * Doing the following when there is already a contact detail in the list will replace it by blank contact details : "Nieuw" -> Don't fill anything -> "Opslaan" (+ it's not persisted, when I refresh the old details are back)

Yes this is normal. All fields in a contact details are optionals + you must click on "Opslaan" (not the one from the modal but the one from the page itself) to have it persisted. If you press the second opslaan, it will be saved and all fields will be empty.
It doesn't seem to be a logical behavior, but I'd prefer to keep it as is for the purpose of the poc, I'll demo it as is to Sofie and see what she thinks of what should we do if an user clicks on new & don't fill anything. Should be fairly easy to adjust for the final implementation.
 
>     * Suggestion - Reuse the buttons formatting of "Bewerk kerngegevens" to make it stand out a little more ?

Design will probably change a lot in the final implementation when I got the feedback of Saska & Sofie 😄 

I've fixed the other comments, thank you !

claire-lovisa (approved):

PR #337, opened 2022-05-09
Fix state not passed to the checkboxes

claire-lovisa:


nbittich:
after looking at it again it's indeed correct that checking huidige positie should empty the date.

nbittich (approved):

PR #344, opened 2022-05-16
Cosmetic change front-end copy mechanism

nbittich:
OP-1389 & last comment of Sofie in OP-1399

### DIFF
--- a/app/components/contact-details.hbs
+++ b/app/components/contact-details.hbs
@@ -32,12 +32,12 @@
     </AuHeading>
 
     <div class="au-o-box au-o-box--small au-c-card au-c-card--fill">
-               <AuButtonGroup class="au-c-button-group--align-right">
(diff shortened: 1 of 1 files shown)


claire-lovisa (changes requested):
When you scroll a bit on the `Bewerk mandataris` page (same for bestuurlid and bedinaar), there is an empty `Contactgegevens` bow, I guess that shouldn't be there ? :sweat_smile:

nbittich:
> When you scroll a bit on the `Bewerk mandataris` page (same for bestuurlid and bedinaar), there is an empty `Contactgegevens` bow, I guess that shouldn't be there ? sweat_smile

I forgot to save I think -_-" I changed according to comments I get from Saska, can you have a look again? thx!!

claire-lovisa (approved):
It's better like that haha :)

PR #346, opened 2022-05-17
Person Search query : names with more than one word give no results (…

nbittich:
…#343)


nbittich:
no need to merge, will release from there

PR #353, opened 2022-05-25
error service

nbittich:


app/controllers/administrative-units/administrative-unit/ministers/new.js
```diff
@@ -15,6 +15,8 @@ export default class AdministrativeUnitsAdministrativeUnitMinistersNewController
   @service router;
   @service store;
   @service contactDetails;
+  @service contactDetails;
```
claire-lovisa, line 18: Extra `contactDetails` ? :grin:

claire-lovisa (approved):

app/controllers/administrative-units/administrative-unit/ministers/new.js
nbittich: oops indeed

PR #358, opened 2022-06-02
fix bug chrome & aumodal

nbittich:
OP-1469

How to reproduce the bug (without the fix):
- make sure you're using Chrome 102
- open the contact details modal (edit a position)
- press either cancel or selecteer button
- the page is blocked & you can't do anything

Happens to all environments (dev, qa)

this is temporary, once the issue is fixed in appuniversum, we should do OP1474

### DIFF
--- a/app/components/contact-details.hbs
+++ b/app/components/contact-details.hbs
@@ -103,7 +103,7 @@
 
   </div>
 
-  <AuModal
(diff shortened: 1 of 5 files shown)


claire-lovisa (approved):
Damn, tricky, well done :)

PR #362, opened 2022-06-08
when open a new tab, we should be redirected to position

nbittich:
OP-1426

### DIFF
--- a/app/routes/people/index.js
+++ b/app/routes/people/index.js
@@ -5,7 +5,6 @@ import { keepLatestTask } from 'ember-concurrency';
 export default class PeopleIndexRoute extends Route {
   @service store;
   @service muSearch;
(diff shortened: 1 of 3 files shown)


app/routes/people/index.js
```diff
@@ -56,5 +56,19 @@ export default class PeopleIndexRoute extends Route {
         return entry;
       },
     });
+
+    for (const person of page.toArray()) {
+      // TODO
+      // another option would be to reindex everything & keeping the type instead
+      // but this seems too much just to allow opening the link in a new tab...
+      // if it is a better option (reindex), we can do it as well
```
claire-lovisa, line 64: I like it like that ! It's still quite simple and clear code :)

claire-lovisa (approved):

PR #366, opened 2022-06-14
Onbard new administrative unit types

claire-lovisa:
OP-1590


app/routes/people/index.js
```diff
@@ -67,6 +68,28 @@ export default class PeopleIndexRoute extends Route {
...
         person.positionRoute = 'people.person.positions.mandatory';
+      } else if (person.uri.includes('/functionarissen/')) {
+        person.positionRoute = 'people.person.positions.agent';
+      }
+
+      // In the case of functionarissen, a bestuursfunctie is linked to multiple units (OCMW and Gemeente)
+      person.organizations = [];
+      if (Array.isArray(person.organization_id)) {
+        let i = 0;
+        person.organization_id.map((id) => {
```
nbittich, line 79: I don't think it should be map here but forEach? Something like this (not tested) seems clearer:

```javascript

        for (let [index, id] of person.organization_id.entries()) {
            person.organizations.push({
            id: id,
            classification: person.organization_classification[index],
          });    
        }

nbittich (approved):
tested and works as described in the FP. small non blocking comment (see below), but the rest looks good! great job! 🥳

app/routes/people/index.js
claire-lovisa: Aha, true that I did something weird over there :joy: I'll fix it, thanks for the suggestion :)

nbittich (approved):

PR #378, opened 2022-08-01
only one gemeente/province with  type betrokkenheid relationship toezicht

nbittich:
OP-1628

### DIFF
--- a/app/controllers/administrative-units/administrative-unit/local-involvements/edit.js
+++ b/app/controllers/administrative-units/administrative-unit/local-involvements/edit.js
@@ -12,6 +12,7 @@ export default class AdministrativeUnitsAdministrativeUnitLocalInvolvementsEditC
   @service router;
   @service store;
   @tracked showTotalFinancingPercentageError = false;
+  @tracked showMoreThanOneFinancialTypeError = false;
   isFinancialInvolvementType = isFinancialInvolvementType;
 
   classificationCodes = [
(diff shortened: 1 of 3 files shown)


app/controllers/administrative-units/administrative-unit/local-involvements/edit.js
```diff
@@ -66,13 +73,21 @@ export default class AdministrativeUnitsAdministrativeUnitLocalInvolvementsEditC
...
+
   @action
   handleInvolvementTypeSelection(involvement, involvementType) {
     involvement.involvementType = involvementType;
 
     if (!involvementType || involvementType.id !== INVOLVEMENT_TYPE.FINANCIAL) {
       involvement.percentage = 0;
     }
+
+    if (this.isLessOrOneFinancialLocalInvolvement) {
```
claire-lovisa, line 88: Shouldn't `isLessOrOneFinancialLocalInvolvement` is `isOneOrLessFinancialLocalInvolvement` ? Pure syntaxic question but "less or one" doesn't make sense no ? :sweat_smile:

claire-lovisa (approved):
Tested OK :heavy_check_mark:

app/controllers/administrative-units/administrative-unit/local-involvements/edit.js
nbittich: Indeed! I renamed it to `isOneOrLessFinancialLocalInvolvement`

PR #379, opened 2022-08-08
leave percentage empty but editable in case of mede financiering

nbittich:
OP-1630

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,5 @@
 .DS_Store
 
 .history
+
+.vscode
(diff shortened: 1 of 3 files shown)


claire-lovisa (changes requested):
When I select `Toezichthoudend` I can edit the percentage but we shouldn't be able if I understand properly the task ?

nbittich:
> When I select `Toezichthoudend` I can edit the percentage but we shouldn't be able if I understand properly the task ?

woops i edited the comment instead of reply <_< 
You should be able to edit mede financering and toezicht but not adviserend I think.

claire-lovisa (approved):
Ah ok I didn't understood the title of the task properly then, it works :)

PR #380, opened 2022-08-10
don't show the clear button if municipality selected

nbittich:


claire-lovisa (approved):
Very nice, small update but improve greatly the UX, thx :)

PR #386, opened 2022-08-17
Fix logic of governing body status component

claire-lovisa:
OP-1655

I also fixed the sorting of the governing bodies

### DIFF
--- a/app/components/governing-body-status.js
+++ b/app/components/governing-body-status.js
@@ -2,13 +2,16 @@ import Component from '@glimmer/component';
 
 export default class GoverningBodyStatusComponent extends Component {
   get isActive() {
-    if (this.args.date) {
-      var today = new Date();
-      if (this.args.date > today) {
+    const today = new Date();
+
+    if (this.args.startDate) {
+      if (!this.args.endDate || this.args.endDate > today) {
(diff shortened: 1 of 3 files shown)


app/components/governing-body-status.js
```diff
@@ -2,13 +2,16 @@ import Component from '@glimmer/component';
...
+    if (this.args.startDate) {
+      if (!this.args.endDate || this.args.endDate > today) {
         return true;
       } else {
         return false;
       }
+    } else if (this.args.endDate) {
+      return this.args.endDate > today;
     }
     return null;
```
nbittich, line 16: maybe this can be simplified with something like
```javascript
 get isActive() {
      const start = this.args.startDate || new Date();
      return !this.args.endDate || this.args.endDate > start;
  }
 ```

nbittich (approved):

app/components/governing-body-status.js
claire-lovisa: I couldn't simplify it that much because we need to know if a start date was provided or not to set the proper status. I still improved a bit the condition, but I prefer to keep it for readability purposes than to shrink it too much :)

PR #391, opened 2022-08-22
nis code

nbittich:
OP-1634


claire-lovisa (changes requested):
The code here works perfectly fine, but some changes are missing I think in the edit view of the core data, because there the NIS code is showing an empty value

claire-lovisa:
I fixed it quickly @nbittich to be able to release it tonight, sorry :sweat_smile:

PR #393, opened 2022-08-26
warning if nor primary site

nbittich:
OP-1697


claire-lovisa (commented):
The code works well, but I find it a bit confusing to have the message displayed for so little time and then that it disappears.
It would maybe be more user friendly if you instantly change "nee" to "ja" and show the message with no timeout ?

claire-lovisa (approved):

PR #406, opened 2022-10-12
small improvements

iurianu:


nbittich (changes requested):
# Must change:
- package-lock.json shouldn't have changed since there was no new dependency added, this is weird. maybe you need to upgrade to node 16 indeed (check with @Windvis to be sure)

# Remarks for the future:
- please use a meaningful name for the branch (otherwise, it might be "improvements" every time)
- please name your branch with a prefix, "feat/xxx", "chore/xxx", etc
- Please do not reuse the same name from a previous PR for your branch

Windvis:
I gave it a quick try and have the same lock file issue if I use Volta, which uses these versions: https://github.com/lblod/frontend-organization-portal/blob/a0b9f32c1c6df6a1007f57dc54e298843fac9082/package.json#L113-L116

`npm install` on the development branch modifies the lock file. @nbittich @claire-lovisa What node / npm versions are you using exactly? Is that docker-ember? Or managed with a different node management tool? I wouldn't expect different versions of npm 8 to yield different lock file changes, but it _does_ seem that way at the moment. 

Chances are the volta config needs to be adjusted to match something else, we just need to find out what exactly 😄.

nbittich:
@Windvis  currently my setup is:
```
{
  npm: '8.15.0',
  node: '16.17.1',
  v8: '9.4.146.26-node.22',
  uv: '1.43.0',
  zlib: '1.2.11',
  brotli: '1.0.9',
  ares: '1.18.1',
  modules: '93',
  nghttp2: '1.47.0',
  napi: '8',
  llhttp: '6.0.9',
  openssl: '1.1.1q+quic',
  cldr: '41.0',
  icu: '71.1',
  tz: '2022a',
  unicode: '14.0',
  ngtcp2: '0.1.0-DEV',
  nghttp3: '0.1.0-DEV'
}
```

nbittich (approved):

Windvis:
@nbittich Even with those versions I still see the changed lock file after an `npm install` on the dev branch. That doesn't happen for you?

nbittich:
Indeed @Windvis , it was in a previous PR, I missed that one:
https://github.com/lblod/frontend-organization-portal/pull/400/files#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519

Windvis:
Ok, check, then it probably makes sense to simply commit the correct version on the dev branch and update the volta config to match your versions (or alternatively the versions used in the docker ember image). The volta config can also be removed if nobody uses it. It seems I added that in the past 😄.

PR #407, opened 2022-10-13
Bug fixes

iurianu:
OP-1856 | OP-1859

### DIFF
--- a/app/styles/components/_environment-banner.scss
+++ b/app/styles/components/_environment-banner.scss
@@ -37,16 +37,23 @@ $au-env-element-height: 4rem;
 .au-c-app--environment {
   position: relative;
   overflow-y: auto;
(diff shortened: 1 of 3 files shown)


nbittich:
Are you sure this implements OP-1839?  it doesn't seem so

iurianu:
> Are you sure this implements OP-1839? it doesn't seem so

The branch was created for Op-1839, and while working on it, I found the bug in OP-1859

claire-lovisa:
So this PR is only for OP-1859 and a new PR will be created for OP-1839 ?

claire-lovisa (approved):
Approving after a discussion during the standup about this PR being only for OP-1859. @iurianu can you change the title and description of this PR so that it really reflects what it's doing ? Thanks ! :)

PR #408, opened 2022-10-17
Fix references to region vs municipatily

bdevloed:
Refer to the region instead of the municipality working area.
The municipality working area is located within the region.

Note that the first iteration uses an implicit reference to the higher level, a possible improvement is to check whether that higher level is a `Referentieregio`.


claire-lovisa (approved):
Works as is for now, with a small risk indeed to one day display the wrong level of information. I'm still approving, if you decide to update it i'll re-review :)

PR #409, opened 2022-10-20
Improve Search Instructions

iurianu:
OP-1839


claire-lovisa:
@iurianu I'll review it when I have a bit of time but in the meantime can you add in the description of the issue which ticket you are fixing ?
Also, for the future, we prefix the branches with `feature/` when it's a feature, `fix/` when it's a bugfix, `/chore` when it's a core update of the app. It helps keeping things organized :smile:

app/components/table-message/no-result.hbs
```diff
@@ -0,0 +1,11 @@
+{{!-- No Result Message --}}
+<TableMessage>
+    <p class="au-u-margin-bottom-large no-result">No result Er werden geen zoekresultaten gevonden. Kijk na op spelfouten, probeer een andere zoekopdracht of voeg een nieuwe persoon toe!</p>
+    <AuLink
+        @route="people.new"
+        @query={{hash redirectUrl=@personCreationRedirectUrl}}
+        class="au-c-button au-c-button--secondary"
+    >
+        + Voeg een nieuwe persoon toe!
```
claire-lovisa, line 9: I think the `!` at the end should not be there ?

app/components/table-message/some-result.hbs
```diff
@@ -0,0 +1,10 @@
+{{!-- Partial Result Message --}}
+<TableMessage>
+    <p class="some-result">
+        Als u de persoon die u zoekt niet vindt, kijk na op spelfouten of probeer een andere zoekopdracht. 
+        Indien de persoon ontbreekt, voeg een nieuwe persoon toe
+        <a href="mailto:{{config "contactEmail"}}" class="au-c-link">
+            + {{config "contactEmail"}}
```
claire-lovisa, line 7: `voeg een nieuwe persoon toe` should be in the link instead of `+ {{config "contactEmail"}}`

tests/integration/components/table-message/no-result-test.js
```diff
@@ -0,0 +1,26 @@
...
+    // Template block usage:
+    await render(hbs`
+      <TableMessage::NoResult>
+        template block text
+      </TableMessage::NoResult>
+    `);
+
+    assert.dom(this.element).hasText('template block text');
+  });
+});
```
claire-lovisa, line 26: The test file can be removed as it doesn't do anything :)

tests/integration/components/table-message/some-result-test.js
```diff
@@ -0,0 +1,26 @@
...
+    // Template block usage:
+    await render(hbs`
+      <TableMessage::SomeResult>
+        template block text
+      </TableMessage::SomeResult>
+    `);
+
+    assert.dom(this.element).hasText('template block text');
+  });
+});
```
claire-lovisa, line 26: Same, can be removed !

app/components/person-search.hbs
```diff
@@ -104,6 +94,12 @@
             </td>
           </c.body>
         {{/if}}
+        {{#if (and
+          (not this.searchPeopleTask.isRunning)
+          (not this.searchPeopleTask.last.isError)
+        )}}
+          <TableMessage::SomeResult />
```
claire-lovisa, line 101: I really like that you made components out of it, I just find `SomeResult`a weird naming, maybe `HasResults`would fit better the use case ? Or something else, but maybe something a bit more precise ?

claire-lovisa (changes requested):
Some changes required and some suggestions, see comments ! Thanks :smile:

app/components/person-search.hbs
```diff
@@ -5,7 +5,7 @@
 >
   <SecuredArea>
     <:edit>
-      <p class="au-u-margin-bottom">
+      <p class="au-u-margin-bottom au-u-hidden">
```
claire-lovisa, line 8: Any reason why it's hidden but not deleted ?

app/components/table-message/no-result.hbs
```diff
@@ -0,0 +1,11 @@
+{{!-- No Result Message --}}
+<TableMessage>
+    <p class="au-u-margin-bottom-large no-result">No result Er werden geen zoekresultaten gevonden. Kijk na op spelfouten, probeer een andere zoekopdracht of voeg een nieuwe persoon toe!</p>
+    <AuLink
+        @route="people.new"
+        @query={{hash redirectUrl=@personCreationRedirectUrl}}
```
claire-lovisa, line 6: The redirection doesn't work as expected here, I think it's because `@personCreationRedirectUrl` is not passed down to the component ?

claire-lovisa (changes requested):

claire-lovisa:
I also noticed than when we click on "Zoek", the table displays "No data" instead of "Aan het laden..." when it's searching for data. I'm not seeing why at first sight, to be investigated :)

app/components/table-message/no-result.hbs
iurianu: @claire-lovisa I checked it, and it redirects to http://localhost:4200/personen/nieuw

claire-lovisa: I mean now, when do create the new person in the flow to create a new position, you are then redirected to `/personen/:id/contactgegevens` instead of the position creation page and that is fixed by calling the component like ```<TableMessage::NoResult @personCreationRedirectUrl={{@personCreationRedirectUrl}}/>```

app/templates/administrative-units/index.hbs
```diff
@@ -140,20 +140,7 @@
...
-              <a
-                class="au-c-link"
-                href="mailto:{{config "contactEmail"}}"
-              >{{config "contactEmail"}}</a>
-              of
-              <AuLink @route="administrative-units.new">maak een organisatie
-                aan.</AuLink>
-            </p>
-          </TableMessage>
+          <TableMessage::NoResult />
```
claire-lovisa, line 143: I think here you should revert it to the previous version, when we are in the administrative unit page it makes no sense to provide the option to create a new person and I don't see it in the designs

app/templates/people/index.hbs
```diff
@@ -112,17 +112,7 @@
...
-              Er werden geen zoekresultaten gevonden. Kijk na op spelfouten, of
-              probeer een andere zoekopdracht. Indien een persoon ontbreekt,
-              meldt dit aan
-              <a
-                href="mailto:{{config 'contactEmail'}}"
-                class="au-c-link"
-              >{{config "contactEmail"}}</a>.
-            </p>
-          </TableMessage>
+          <TableMessage::NoResult />
```
claire-lovisa, line 115: Similar remark, I don't see this change in the designs and I'm afraid that doing this change will encourage the users to create persons that might already exist but who just don't have a position attached to them yet, which is the opposite of what we are trying to achieve with this ticket no ?

claire-lovisa:
> > I also noticed than when we click on "Zoek", the table displays "No data" instead of "Aan het laden..." when it's searching for data. I'm not seeing why at first sight, to be investigated :)
> 
> I can't find it anywhere inside the app. Maybe it's a default from PowerSelect

The loading message comes from [here](https://github.com/appuniversum/ember-appuniversum/blob/development/addon/components/au-data-table/content-body.hbs#L2) , it just means that when we do ```      @isLoading={{and
        (not this.searchPeopleTask.lastSuccessful.value)
        (this.searchPeopleTask.isRunning)
      }}``` to notify the component if data is loading or not, the value isn't what it should be

claire-lovisa:
> > > I also noticed than when we click on "Zoek", the table displays "No data" instead of "Aan het laden..." when it's searching for data. I'm not seeing why at first sight, to be investigated :)
> > 
> > 
> > I can't find it anywhere inside the app. Maybe it's a default from PowerSelect
> 
> The loading message comes from [here](https://github.com/appuniversum/ember-appuniversum/blob/development/addon/components/au-data-table/content-body.hbs#L2) , it just means that when we do
> 
> ```
>       @isLoading={{and
>         (not this.searchPeopleTask.lastSuccessful.value)
>         (this.searchPeopleTask.isRunning)
>       }}
> ```
> 
> to notify the component if data is loading or not, the value isn't what it should be

Ha, found it, the parenthesis around `(this.searchPeopleTask.isRunning)` apparently made it fail. I pushed the fix as it was small

app/components/table-message/has-results.hbs
```diff
@@ -0,0 +1,10 @@
+{{!-- Partial Result Message --}}
+<TableMessage>
+    <p class="has-results">
+        Als u de persoon die u zoekt niet vindt, kijk na op spelfouten of probeer een andere zoekopdracht. 
+        Indien de persoon ontbreekt, 
+        <a href="mailto:{{config "contactEmail"}}" class="au-c-link">
+            + voeg een nieuwe persoon toe
+        </a>.
```
claire-lovisa, line 8: Here, should we really be sending an email ? Don't we want to just set a link to the `person/new` page ..?

claire-lovisa (changes requested):

claire-lovisa (approved):
I think we're good ! Thanks for the work !

PR #417, opened 2022-11-15
On creating a new person follow up with linking it to a position

iurianu:
OP-1893


claire-lovisa (changes requested):
This doesn't work for all the cases. When you create a position via selecting a bestuurseenheid, then Bedienaren -> Nieuw , the redirection breaks because you don't have any `?` indicating the query params

claire-lovisa (changes requested):
Have you tested the code ? Because on my side it doesn't work at all when creating positions from the Bedienaren or the Bestuursorganen pages ... I think you inverted the `?` and `&` :thinking: 

Locally I tested with `this.redirectUrl.lastIndexOf('?') == -1 ? '?' : '&'` which seemed to work

claire-lovisa (approved):

PR #420, opened 2022-12-13
Implement Environment Banner

iurianu:
OP-1927 » OP-1965


iurianu:
The changes added in the first commit were made in 2 different tasks - change button order, and fix double scroll!

nbittich:
On my machine it looks like this:
Chrome:
image

Firefox:
image


It seems that the header "Vlaanderen is het OrganisatiePortaal" is not shown properly?

Same when logged in:
image

It seems that the name is no longer visible.

nbittich (changes requested):
it doesn't seem to be displayed properly locally on my machine (see screenshots)

package.json
```diff
@@ -27,12 +27,14 @@
...
   "devDependencies": {
+    "@appuniversum/appuniversum": "^1.2.0",
     "@appuniversum/ember-appuniversum": "^1.1.1",
     "@ember/optional-features": "^2.0.0",
     "@ember/test-helpers": "^2.6.0",
     "@glimmer/component": "^1.0.4",
     "@glimmer/tracking": "^1.0.4",
-    "@lblod/ember-acmidm-login": "^1.4.0",
+    "@lblod/ember-acmidm-login": "^1.4.0",    
+    "@lblod/ember-environment-banner": "git://github.com/lblod/ember-environment-banner.git#env-banner-op",
```
claire-lovisa, line 37: Here usually we don't commit that, we wait until the package is released with our changes and we bump it

app/controllers/application.js
```diff
@@ -1,29 +1,32 @@
...
+  get isLocalhost() {
+    if (
+      window.location.hostname === 'localhost' ||
+      window.location.hostname === '[::1]'
+    ) {
+      return true;
+    } else {
+      return false;
+    }
+  }
```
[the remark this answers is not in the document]
claire-lovisa: Il would also make testing easier, to see how the banner behaves depending on which param we give. Now on localhost it always gets overriden by the localhost value, but I'd be convenient to be able to view the other perspectives of the other envs !

app/controllers/application.js
```diff
@@ -1,29 +1,32 @@
...
 
-const isLocalhost = Boolean(
-  window.location.hostname === 'localhost' ||
-    // [::1] is the IPv6 localhost address.
-    window.location.hostname === '[::1]'
-);
+  get environment() {
+    return getOwner(this).resolveRegistration('config:environment').environment;
+  }
```
claire-lovisa, line 17: This is not used anymore no ?

app/controllers/application.js
```diff
@@ -1,29 +1,32 @@
...
 
-  get environmentTitle() {
-    const thisEnvironmentValues = isLocalhost
-      ? 'lokalomgeving'
-      : getOwner(this).resolveRegistration('config:environment')
-          .environmentTitle;
-
-    return thisEnvironmentValues;
+  get applicationName() {
+    return getOwner(this).resolveRegistration('config:environment').appName;
```
claire-lovisa, line 29: This is not used anymore no ?

claire-lovisa:
Looks good ! :) I'm just waiting for the addon bump to approve it

claire-lovisa:
I'm having second thoughts about it. I'm the sure the ember environment detection is reliable enough. I'm not sure it'll be able to detect in which env we are, because for example if you go to dev and use the ember inspector, you can see that the environment detected seems to be prod, even though it's not the case
image

So we probably should go as GN, use environment variables in the docker-compose.override.yml files of dev and qa to set the environment name.

claire-lovisa:
Indeed, I checked, all built versions of the frontend are built via the `production` configuration so we can't rely on this. So I think the best would be to rely on the `EMBER_ENVIRONMENT_NAME` env variable on the frontend as GN does

claire-lovisa (approved):

PR #421, opened 2022-12-13
Fix - Feat/Redo change button order

iurianu:
* [OP-1966](https://binnenland.atlassian.net/browse/OP-1966)
* Redo changes from [OP-1757](https://binnenland.atlassian.net/browse/OP-1757), required within [OP-1927](https://binnenland.atlassian.net/browse/OP-1927)

### DIFF
--- a/app/styles/_shame.scss
+++ b/app/styles/_shame.scss
@@ -123,7 +123,7 @@ table.dataTable tbody th {
 // Reverse Button Order :: OP-1757
 .au-c-button-group--align-right {
   @include mq($from:xsmall) {
-    justify-content: flex-start;
+    justify-content: flex-end;
   }
 }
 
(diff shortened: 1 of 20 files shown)


app/styles/_shame.scss
```diff
@@ -120,9 +120,11 @@ table.dataTable tbody th {
 }
 
 // Allow buttons to be aligned on the right side
+// Reverse Button Order :: OP-1757
 .au-c-button-group--align-right {
-  @include mq($from:medium) {
-    justify-content: flex-end;
+  @include mq($from:xsmall) {
+    justify-content: flex-start;
+    flex-direction: row-reverse;
```
claire-lovisa, line 127: This can now get back to how it was before no ?

claire-lovisa (changes requested):
The CSS can be reverted, and the order of `Bewerk bestuursorgaan` still needs a change. But after that I think we're good !

PR #423, opened 2022-12-19
Fix/Align Labels to Baseline

iurianu:
* [OP-1967](https://binnenland.atlassian.net/browse/OP-1967) 
* Requirements from [OP-848](https://binnenland.atlassian.net/browse/OP-848) - Form labels are always centered vertically


nbittich (approved):
works as expected

PR #424, opened 2022-12-21
hotfix/fix container height

iurianu:


bdevloed (approved):
I'm not a CSS expert, but I tested the result and it indeed looks again like before.

PR #425, opened 2022-12-29
Change date format

iurianu:
[OP-1983](https://binnenland.atlassian.net/browse/OP-1983) - Change date output format to DD-MM-YYYY


nbittich (changes requested):
I don't understand why not just modifying `app/helpers/date-format.js`, it seems to provide the same result without changing 22+ files.

iurianu:
> I don't understand why not just modifying `app/helpers/date-format.js`, it seems to provide the same result without changing 22+ files.

I missed the helper file - I guess I still have a lot to learn on ember :/
Thanks!

nbittich (approved):

PR #428, opened 2023-01-03
Feature/fix missing persons

nbittich:
OP-1987

requires https://github.com/lblod/app-organization-portal/pull/207


nbittich:
bad branch

PR #432, opened 2023-01-10
Disable position creation or edition from 01/02/2023

claire-lovisa:
OP-2032

This will not be testable by the testers, what I did to test locally was to update the `new Date('2023-02-01')` by today's date.
As it'll not be tested by the testers, an extensive testing of this one would be valuable :smile: Thanks !


nbittich (approved):

bdevloed (approved):
I tested this by changing the date as indicated, the app behaved as expected. No issues found :)

PR #436, opened 2023-01-11
Fix data table navigation in bestuurseenheid page

claire-lovisa:
OP-2016

I also changed the way the municipality select worked, because when we selected a municipality and then refreshed, the municipality filter was empty.


bdevloed (commented):
Fixed indeed :tada:

bdevloed (approved):

PR #437, opened 2023-01-12
Make all pages scrollable on small screens

claire-lovisa:
OP-2028


bdevloed (approved):
Works and looks much cleaner now.

PR #439, opened 2023-01-18
Fix filtering organen

claire-lovisa:
OP-2000


nbittich (approved):
works as expected

bdevloed (approved):
Works with chrome too now :)

PR #440, opened 2023-01-18
Feat / Change Headline in Templates

iurianu:
OP-1982 - Related organisations: change titel (h3) in overview of district, gemeente and OCMW


claire-lovisa:
(hey, small note before i review, we only put the task number but not the link for confidentiality reasons)

claire-lovisa (changes requested):
The ticket and PR mentions that the change should only happen for district, gemeente and OCMW, but it happens for all the types now ?

claire-lovisa (changes requested):
The change for districts should be kept

claire-lovisa (approved):

PR #443, opened 2023-01-25
draft hard split2023

nbittich:
covers roughly
 OP-2025


app/services/contact-details.js
```diff
@@ -116,8 +116,12 @@ export default class ContactDetailsService extends Service {
     const positions = [];
 
     const mandatories = (await person.mandatories).toArray(); // mandatarissen
+    console.log(mandatories.length);
```
claire-lovisa, line 119: Can they be removed ? :)

PR #445, opened 2023-02-08
Update link in banner

claire-lovisa:
OP-2133


nbittich (approved):
maybe this should also be changed (as from my understanding we disabled editing person)
Bewerk = edit I think
image

claire-lovisa:
good remark, i'll ask Sofie

claire-lovisa:
but that'll be for a different PR :)

PR #453, opened 2023-03-08
validate involvement type

nbittich:
OP-2220, had to fight against ember proxies all over the place. doing a simple  change requires a Phd 😢 , but hopefully it seems working. if you see a better way please tell me because I couldn't do it otherwise


claire-lovisa (approved):
I think the code looks good, even though you had a hard time writing it, I find it clear now when reading it ! :sweat_smile:

PR #459, opened 2023-05-15
governing bodies date validation

nbittich:
OP-2356


µ
```diff
@@ -0,0 +1,76 @@
+import Controller from '@ember/controller';
```
claire-lovisa, line 1: What is that file ? :sweat_smile:

app/validations/governing-body.js
```diff
@@ -0,0 +1,85 @@
...
+        return 'Kies een startdatum die vóór de einddatum plaatsvindt';
+      }
+      return validateNoOverlap(governingBodies, newStartDate, effectiveEndDate);
+    }
+
+    return true;
+  };
+}
+
+function validateNoOverlap(otherBodies, startDate, endDate) {
```
claire-lovisa, line 60: For the overlap, should we compare in strictly instead of `>=` / `<=` ? I'm thinking about all the organen we got from Loket, it's very common that one stops first of jan and the second one starts the same day. As they are datetimes, it's (I think) always at midnight, so it's probably safe to use strict comparison ?

nbittich: indeed! I've fixed it

µ
nbittich: vim going crazy 😬

claire-lovisa (approved):

PR #474, opened 2023-06-30
Fix sorting in related organizations tab

claire-lovisa:
OP-2439


bdevloed:
It works, but I found an unexpected glitch:
when clicking the sort button a second time, before the table is updates. I get a page not found.

Not sure whether it's something that has to be fixed right now though - to be checked with Sofie

claire-lovisa:
@bdevloed I found a trick to go around the bug :sweat_smile: it's how we do it in other pages as well.

bdevloed (approved):

PR #475, opened 2023-07-04
Feat/igs add new

lagartoverde:
Solves this https://binnenland.atlassian.net/browse/OP-2521, pending answer from Sofie about the Doel field


app/validations/administrative-unit.js
```diff
@@ -98,6 +98,20 @@ export default {
...
+    [
+      validatePresence({
+        presence: true,
+        ignoreBlank: true,
+        message: 'Vul de datum in',
+      }),
+      validateFutureDate,
+    ],
+    function (changes, content) {
+      console.log('running');
```
claire-lovisa, line 111: `running` :grin:

app/templates/administrative-units/new.hbs
```diff
@@ -146,6 +146,23 @@
...
+                @errorMessage={{@model.administrativeUnitChangeset.error.geplandeEinddatum.validation}}
+              >
+                <:label>Geplande einddatum</:label>
+                <:content  as |hasError|>
+                  <AuDatePicker 
+                    @onChange={{fn
+                      (mut @model.administrativeUnitChangeset.geplandeEinddatum)
+                    }}
+                    @error={{hasError}}
+                  />
```
claire-lovisa, line 162: I do like this component a lot, but in OP we tend to use `<Datepicker />`. I think that it would be a little weird for consistency not to use the name kind of datepicker accross the application, but might be worth asking Sofie which one she prefers.

If I remember correctly, we had to use a custom component for having a min / max on the dates. We used them when creating new persons or positions, but currently it's Loket which is in charge of creating those resources so we can probably safely change it. Might be worth to make a separate ticket for it though :grin:

claire-lovisa (commented):
I see you adapted fast to OP hehe :) Just a remark to double check about the date picker, but the code works well !

claire-lovisa (approved):

PR #476, opened 2023-07-05
Feature/search page filters

lagartoverde:
Solves https://binnenland.atlassian.net/browse/OP-2516

### DIFF
--- a/app/controllers/administrative-units/index.js
+++ b/app/controllers/administrative-units/index.js
@@ -61,7 +61,7 @@ export default class AdministrativeUnitsIndexController extends Controller {
   }
 
   get modelHasOnlyWorshipAdministrativeUnits() {
-    if (this.administrativeUnits) {
+    if (this.administrativeUnits && !this.isIGS) {
       return !this.administrativeUnits.toArray().some((adminUnit) => {
         return (
           adminUnit.classification_id !== CLASSIFICATION_CODE.WORSHIP_SERVICE &&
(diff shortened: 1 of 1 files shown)


claire-lovisa (commented):
In the end, is the condition really necessary ? I think I have the same result without being on the branch :sweat_smile:

claire-lovisa (changes requested):
So I think that we don't have to check `!this.isIGS` so that part can probably be removed, but I did find an issue. When you type the name of an IGS and try to filter on that, it doesn't work UNLESS you also picked its classification. It comes from the route model of `administrative-units/index.js` where we do
```
    if (params.classificationId) {
      filter['classification_id'] = params.classificationId;
    } else {
      if (this.currentSession.hasWorshipRole) {
        filter['classification_id'] = `
         ${CLASSIFICATION.CENTRAL_WORSHIP_SERVICE.id},
         ${CLASSIFICATION.WORSHIP_SERVICE.id},
        `;
      } else {
        filter['classification_id'] = `
         ${CLASSIFICATION.AGB.id},
         ${CLASSIFICATION.APB.id},
         ${CLASSIFICATION.MUNICIPALITY.id}
         ${CLASSIFICATION.PROVINCE.id}
         ${CLASSIFICATION.OCMW.id}
         ${CLASSIFICATION.DISTRICT.id}
       `;
      }
    }
```

We need to add the new classifications here as well, in the else block

PR #477, opened 2023-07-05
Change events - IGS

claire-lovisa:
OP-2514


lagartoverde (approved):
I was thinking, that maybe we should export the check of isIGS to an utility function, as we are repeating that piece of code everywhere

claire-lovisa:
We could, but I'm not sure it would be that easy cause we have slightly different ways of checking it depending on how the model is structured etc :thinking:

lagartoverde:
Yes, I was thinking more like `isIGS(classificationId)`

claire-lovisa:
Feel free to add it :grin:

PR #478, opened 2023-07-05
Hide organen in menu

claire-lovisa:
OP-2523

### DIFF
--- a/app/templates/administrative-units/administrative-unit.hbs
+++ b/app/templates/administrative-units/administrative-unit.hbs
@@ -30,7 +30,7 @@
             </AuNavigationNestedLink>
           </li>
         {{/if}}
-        {{#unless this.isAgbOrApb}}
+        {{#if (and (not this.isAgbOrApb) (not this.isIgs))}}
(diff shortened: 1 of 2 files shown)


app/templates/administrative-units/administrative-unit.hbs
```diff
@@ -30,15 +30,15 @@
             </AuNavigationNestedLink>
           </li>
         {{/if}}
-        {{#unless this.isAgbOrApb}}
+        {{#if (and (not this.isAgbOrApb) (not this.isIgs))}}
```
lagartoverde, line 33: What's the reason of using the if here instead of the `unless (or this.isAgbOrApb this.isIgs)

lagartoverde (approved):
Looks good, just a minor comment, but it can be merged

app/templates/administrative-units/administrative-unit.hbs
claire-lovisa: Ember complains if you put a `not` in an `unless` :sweat_smile: I don't remember the exact warning message though.

PR #479, opened 2023-07-05
Feat/vestigingen

lagartoverde:
OP-2512


claire-lovisa (changes requested):
Nice :D I think that we just miss the fact that when we create a new IGS, the site type doesn't get set. I think that it happens is `controllers/administrative-units/new.js` when doing 
```
      if (this.isNewAgb || this.isNewApb) {
        primarySite.siteType = siteTypes.find(
          (t) => t.id === 'f1381723dec42c0b6ba6492e41d6f5dd'
        );
      }
```

lagartoverde:
Done 👍 thanks

claire-lovisa (approved):

PR #481, opened 2023-07-06
Allow agb and apb in personen filters

claire-lovisa:
OP-2525

### DIFF
--- a/app/templates/people/index.hbs
+++ b/app/templates/people/index.hbs
@@ -136,7 +136,11 @@
             </td>
             <td>
               {{#each person.organizations as |organization index|}}
-                {{#if organization.id}}
+                {{#if (and
+                  organization.id
+                  (not-eq organization.classification "Autonoom gemeentebedrijf")
+                  (not-eq organization.classification "Autonoom provinciebedrijf")
+                )}}
                   <AuLink
                     @route="administrative-units.administrative-unit"
                     @model={{organization.id}}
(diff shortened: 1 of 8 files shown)


app/templates/people/index.hbs
```diff
@@ -136,7 +136,11 @@
             </td>
             <td>
               {{#each person.organizations as |organization index|}}
-                {{#if organization.id}}
+                {{#if (and
```
nbittich, line 139: maybe move this in the controller?

app/templates/people/index.hbs
```diff
@@ -160,6 +164,11 @@
                       ({{organization.classification}})
                     {{/if}}
                   </AuLink>
+                {{else if (or
```
nbittich, line 167: maybe move this in the controller?

nbittich (commented):
I didn't test it yet but from what I see it looks good!

app/templates/people/index.hbs
```diff
@@ -136,7 +136,11 @@
             </td>
             <td>
               {{#each person.organizations as |organization index|}}
-                {{#if organization.id}}
+                {{#if (and
+                  organization.id
+                  (not-eq organization.classification "Autonoom gemeentebedrijf")
```
nbittich, line 141: this part *

claire-lovisa: Initially didn't do it because of the imbrication of `{{#each person.organizations as |organization index|}}` :grimacing: but you're right that id would make more sense, I'll have to create a component for it then I think

nbittich (approved):

PR #486, opened 2023-07-13
Fix validations too strict on municipalities

claire-lovisa:
A validation checking if a municipality had a `wasFoundedByOrganization` relation (that never exists for this type of admin unit) prevented the user to save a municipality. This PR fixes it.

OP-2555


lagartoverde (approved):
Looks good

PR #487, opened 2023-07-20
Set regio as read-only for municipalities

claire-lovisa:
OP-2575


lagartoverde (approved):
Looks good to me, we will have also to change the workaround I did for the IGS to show the regio as read only, as it's no longer needed

PR #501, opened 2023-08-28
Disable validation on planned end date

claire-lovisa:
OP-2668


lagartoverde:
Even if we disable that is a required field, shouldn't we keep the validateFutureDate, so in case any date gets input it checks it's not a past date?

claire-lovisa:
> Even if we disable that is a required field, shouldn't we keep the validateFutureDate, so in case any date gets input it checks it's not a past date?

Ooooh yes you're completely right !! Gonna fix it

claire-lovisa:
@lagartoverde Ready to be re-reviewed :)

lagartoverde (approved):
Looks good to me :)

PR #502, opened 2023-08-29
Solve problem with duplication of sites when changing primary one

lagartoverde:


claire-lovisa (approved):
Seems to be working well :) Two questions related to the ticket:
- Should it be hotfixed to prod?
- From what I can see, the issue was not only frontend but also backend. The issue seems to cause the same site to be saved at the same time as a primary and as a non-primary site. The frontend fix only will not be enough to fix data that potentially got wrong because of the bug ? So we might also need a migration to fix them ?

PR #503, opened 2023-09-01
Refactoring of related organizations model

claire-lovisa:
OP-2581

### Status

:warning: Do not merge as-is, many issues remain. Some purely technical, some others require more refinement :warning: 
To be used against backend https://github.com/lblod/app-organization-portal/pull/321.

### Issue encountered

**Direction of relationships**

In the feature passeport, we talk about showing in the same table relationships that are of different directions. For example, when viewing Province Antwerpen, we want to show in the same table:
- Antwerpen -> founding member -> AGB x
- Antwerpen -> related relationship -> Gemeente y

Data-wise this causes an issue. Each relationship is modeled as a membership. A membership has two main relationships: a member and an organization. The member is related|part of the organization. So in the database, what gets stored is different from what we are trying to display:
- Antwerpen -> founding member -> AGB x
- Gemeente y -> related relationship -> Antwwerpen (because a municipality is part of a province, not the opposite)

The issue it brings is that to be able to display both directions in the same table, we would have to drop the pagination and the sorting. It gets even more complex for Provinces where we also want to display the Districts, via `District z -> Gemeente y -> related relationship -> Antwwerpen`.

A solution that will have to be discussed design-wise is to update the table to something like follows, to be able to always show the relationships in the same direction:

| Type member organization | Member organisation | Type relatie | Type organization | Organization |
|-------------------------------------|------------------------------|----------------|----------------|----------------|
| Province | Antwerpen | founding member of | AGB | X |
| Gemeente | Y | related member of | Province | Antwerpen |

This solution means that we can keep the sorting/pagination. It doesn't fix showing the districts for provinces though.

Other solutions can be discussed, like using multiple tables for each direction on relationships that we could put in tabs, ...

**Implications of changing the model**

Changing the related organizations model not only impacts the related organizations module but many more features:
- The creation of a new admin unit
- The core data
- The validations of administrative units
- Conditional filters in the main pages
- 
--> The scope of this task is much bigger than anticipated

### TODOs

Even though the current draft PR brings a basis for the changes, we still have a long way to go to be able to finish it:
- A lot of the previous relationships remain. You can search for the following keywords to find them:
  `is-sub-organization-of`, `isSubOrganizationOf`, `has-sub-organization`, `subOrganizations`, `is-associated-with`, `isAssociatedWith`, `founded-organizations`, `foundedOrganizations`, `was-founded-by-organization`, `wasFoundedByOrganization`, `participates-in`, `participatesIn`, `has-participants`, `hasParticipants`.
  You can find them in the related organizations components, where they will have to be replaced by their membership equivalents once we have a better view on how to display the tables.
  They are used for the conditional filters too (for example in the /bestuurseenheden page, when a province is selected, only municipalities of this province should appear). For this, I think that we can drop those relationships in favor of the werkingsgebieden hierarchy to build the conditional filters. It can be a story by itself, to reduce slowly the scope of the current PR.
  You can also find them at other random places, they need to be tacled as well. I didn't dig in it yet
- Update the validations related to related organizations
- Tacle the numerous TODOs scattered around the PR
- Test extensively each related organizations component
- Fix sorting on "Type organisatie"


mirdono:
Superseded by #632

PR #505, opened 2023-09-13
change: removed option fusie from change events when creating new cha…

Dachamelemon:
…nge event. #REF: OP-2664


app/models/change-event-type.js
```diff
@@ -39,13 +39,13 @@ export const CHANGE_EVENTS_CENTRAL_WORSHIP_SERVICE = [
 
 export const CHANGE_EVENTS_MUNICIPALITY = [
   CHANGE_EVENT_TYPE.NAME_CHANGE,
-  CHANGE_EVENT_TYPE.FUSIE,
+  //CHANGE_EVENT_TYPE.FUSIE,
```
claire-lovisa, line 42: You can entirely remove the line as it's not needed anymore :)

app/models/change-event-type.js
```diff
@@ -39,13 +39,13 @@ export const CHANGE_EVENTS_CENTRAL_WORSHIP_SERVICE = [
...
   CHANGE_EVENT_TYPE.CITY,
 ];
 
 export const CHANGE_EVENTS_OCMW = [
   CHANGE_EVENT_TYPE.NAME_CHANGE,
-  CHANGE_EVENT_TYPE.FUSIE,
+  //CHANGE_EVENT_TYPE.FUSIE,
```
claire-lovisa, line 48: Same here

app/models/change-event-type.js
```diff
@@ -63,7 +63,7 @@ export const CHANGE_EVENTS_IGS = [
   CHANGE_EVENT_TYPE.IN_ONTBINDING,
   CHANGE_EVENT_TYPE.IN_VEREFFENING,
   CHANGE_EVENT_TYPE.ONTBONDEN_EN_VEREFFEND,
-  CHANGE_EVENT_TYPE.FUSIE,
+  //CHANGE_EVENT_TYPE.FUSIE,
```
claire-lovisa, line 66: I think we want to keep this one. The ticket only mentions removing the option for municipalities and OCMW. This block configures IGS, which is an other type of organization, and as far as I know we want to keep the fusion option for them, at least for now !

claire-lovisa (changes requested):
Some little changes to make it perfect :grin:

claire-lovisa (approved):

PR #507, opened 2023-09-14
Onboarding PZ/HPZ

claire-lovisa:
OP-2696, OP-2697, OP-2698, OP-2699, OP-2700, OP-2701

To be tested with https://github.com/lblod/app-organization-portal/pull/330


lagartoverde (approved):
Looks good 👍

PR #509, opened 2023-09-18
Hide calendar icon in datepicker

lagartoverde:


claire-lovisa (commented):
Seems to be working but I'm noticing a strange behavior, when I enter a date, I type "121220" and then it auto fills in the end of the year to "23" ? Do you have the same or know how to prevent that ? Feels strange

lagartoverde:
> Seems to be working but I'm noticing a strange behavior, when I enter a date, I type "121220" and then it auto fills in the end of the year to "23" ? Do you have the same or know how to prevent that ? Feels strange

It was a option of the date-input, I removed it

claire-lovisa (approved):

PR #512, opened 2023-09-21
Added new classification identifiers to select by name

mirdono:
The classification identifiers for the new kind of units were missing, causing them not to show up as completion suggestions when searching by name.


claire-lovisa (approved):
Very nice catch thanks a lot! :smile:

PR #513, opened 2023-09-22
Moved duplicate classification identifier selection to util

mirdono:
As a follow-up to #512 I moved the duplicated code to a util to prevent similar oversights in the future.

### DIFF
--- a/app/components/administrative-unit-select-by-name.js
+++ b/app/components/administrative-unit-select-by-name.js
@@ -16,7 +16,7 @@ export default class AdministrativeUnitSelectByNameComponent extends Component {
     }
 
     filter['classification_id'] = getClassificationIds(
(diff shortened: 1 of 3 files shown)


mirdono:
Sidenote, based on quick test on the dev data these additional classification filters do not seem to impact the results. It might be possible to remove this altogether.

claire-lovisa:
> Sidenote, based on quick test on the dev data these additional classification filters do not seem to impact the results. It might be possible to remove this altogether.

It does have an impact on the search filter on the main `bestuurseenheden` page. If you enter the name of an administrative unit of the classifications you added, we can now directly select their full name as an option of the power select. Otherwise it was not showing in the options list. So it was indeed a bug :)

claire-lovisa (approved):

PR #516, opened 2023-10-03
Feat/implement rich text editor

lagartoverde:
OP-2622


package.json
```diff
@@ -115,5 +115,8 @@
   "volta": {
     "node": "16.13.2",
     "npm": "8.3.1"
+  },
+  "dependencies": {
```
claire-lovisa, line 119: Any reason it's not with all the others deps in devDependencies ? I'm not sure about the consequences of it though

claire-lovisa (changes requested):
Hi ! From what I can see it's working well, except the button to add an hyeprlink that triggers an error in the console so it probably needs some attention.
I don't know the (recent) editor very well so I don't really have ideas of how things could be done differently, but I'm glad it works as is haha !

claire-lovisa (approved):
:sparkle:

PR #517, opened 2023-10-04
Feature: filter administrative units based on identifier

mirdono:
OP-2725


app/components/administrative-unit-select-by-identifier.js
```diff
@@ -0,0 +1,51 @@
...
+    const result = yield this.muSearch.search({
+      index: 'units',
+      sort: 'name',
+      page: '0',
+      size: '100',
+      filters: filter,
+      dataMapping: (data) => {
+        const entry = data.attributes.identifier;
+
+        // Note: Dev contains data for which the resulting identifiers is not
```
claire-lovisa, line 35: Might also be the case on prod no ? I could imagine some admin units having only one identifier

claire-lovisa (approved):
Works smoothly :smile: and the code is very clear :ok_hand:

app/components/administrative-unit-select-by-identifier.js
mirdono: You are probably right, I rephrased the note.

PR #521, opened 2023-10-06
Feature/apb province municipality select filter

Dachamelemon:


Dachamelemon:
I tried to fix conflicts before merging but i seem to have made a mistake somewhere you can ignore this prequest

app/components/administrative-unit-select.hbs
```diff
@@ -1,7 +1,7 @@
 <div class={{if @error "ember-power-select--error"}} ...attributes>
   <PowerSelect
     @disabled={{if @disabled true false}}
-    @allowClear={{@allowClear}}
+    @allowClear={{not @selectedProvince}}
```
claire-lovisa, line 4: I'm not sure we should enforce that we can clear this component only if there is a selected province. It forces the user of the component to use it together with the province component, which might not be relevant for all the use cases. I think that the responsibility to define whether we should allow the clearing is in the hands of the template/controller calling the component instead

app/components/administrative-unit-select.js
```diff
@@ -1,18 +1,18 @@
...
 import { action } from '@ember/object';
 export default class AdministrativeUnitSelectComponent extends Component {
   @service store;
   @tracked loadedRecord;
 
-  constructor() {
-    super(...arguments);
-    this.loadRecord.perform();
-  }
+  munipalities = trackedTask(this, this.loadAdministrativeUnitsTask, () => [
```
claire-lovisa, line 13: Typo in `munipalities` :grin:

app/components/province-organization-select.hbs
```diff
@@ -1,16 +1,17 @@
 <div class={{if @error "ember-power-select--error"}}>
   <PowerSelect
     @disabled={{if @disabled true false}}
-    @allowClear={{true}}
+    @allowClear={{not @selectedMunicipality}}
```
claire-lovisa, line 4: Same as for the admin unit component, I think the component should not handle this but the caller should instead

app/templates/administrative-units/new.hbs
```diff
@@ -163,7 +163,8 @@
               <Item @labelFor="purpose">
                 <:label>Doel</:label>
                 <:content>
-                  <RichTextEditor
+                  <AuTextarea
```
claire-lovisa, line 166: I guess it's an error due to the merge conflict, but we do want to use the `RichTextEditor` component here

app/templates/administrative-units/administrative-unit/core-data/edit.hbs
```diff
@@ -187,10 +187,9 @@
                   <Item @labelFor="purpose">
                     <:label>Doel</:label>
                     <:content>
-                      <RichTextEditor
-                        @onChange={{fn
-                          (mut @model.administrativeUnit.purpose)
-                        }}
+                      <AuTextarea
```
claire-lovisa, line 190: I guess it's an error due to the merge conflict, but we do want to use the `RichTextEditor` component here

claire-lovisa (changes requested):
We are also entirely missing the functionality in the `Related organizations` tab of APBs, they can also be edited there, can you add it ? :)

app/templates/administrative-units/administrative-unit/core-data/edit.hbs
```diff
@@ -188,9 +188,8 @@
                     <:label>Doel</:label>
                     <:content>
                       <RichTextEditor
-                        @onChange={{fn
-                          (mut @model.administrativeUnit.purpose)
-                        }}
+                        @width="block"
```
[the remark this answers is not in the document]
claire-lovisa: Still has to be rollbacked :)

app/templates/administrative-units/new.hbs
```diff
@@ -164,6 +164,7 @@
                 <:label>Doel</:label>
                 <:content>
                   <RichTextEditor
+                    @width="block"
```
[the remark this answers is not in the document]
claire-lovisa: same !

claire-lovisa (changes requested):
There are still some behaviours that are broken / changed from the default behaviour of the feature:

- Bestuurseenheden page:
  - broke the auto-select of a province when a gemeente is picked up
  - allowClear on the province should be false when a muncipality is selected
  
Core data, edit page:
  - allowClear on the province should be false when a muncipality is selected
  
Related organizations, edit page:
  - autoselect of a province when a gemeente is picked up doesn't work
  - when a municipality is picked up, we can still chose from all the provinces instead of having only one choice
  - allowClear on the province should be false when a muncipality is selected

claire-lovisa:
I'm only seeing it now but this behavior should also be added to the `new` route, when a user created a new administrative unit.

claire-lovisa (approved):
Thank you :smile:

PR #523, opened 2023-10-17
Feature/organisationtype multiselect filter

Dachamelemon:


app/templates/administrative-units/index.hbs
```diff
@@ -44,8 +44,8 @@
           </div>
           <div class="au-o-grid__item au-u-1-2@small au-u-1-1@medium">
             <AuLabel for="classification">Type bestuur</AuLabel>
-            <ClassificationSelect
-              @selected={{this.classificationId}}
+            <ClassificationMultiSelect
```
claire-lovisa, line 47: To keep the components naming consistent between multi-selects, would it be possible to rename the component to `ClassificationMultipleSelect`?

app/components/classification-multi-select.js
```diff
@@ -0,0 +1,104 @@
...
+        CLASSIFICATION_CODE.DISTRICT,
+        CLASSIFICATION_CODE.AGB,
+        CLASSIFICATION_CODE.APB,
+        CLASSIFICATION_CODE.PROJECTVERENIGING,
+        CLASSIFICATION_CODE.DIENSTVERLENENDE_VERENIGING,
+        CLASSIFICATION_CODE.OPDRACHTHOUDENDE_VERENIGING,
+        CLASSIFICATION_CODE.OPDRACHTHOUDENDE_VERENIGING_MET_PRIVATE_DEELNAME,
+        CLASSIFICATION_CODE.POLICE_ZONE,
+        CLASSIFICATION_CODE.ASSISTANCE_ZONE,
+      ];
```
claire-lovisa, line 60: This part is not relevant here as it's only used on creation of a new admin unit, which is in the normal `ClassificationSelect` component

app/controllers/administrative-units/index.js
```diff
@@ -27,6 +27,7 @@ export default class AdministrativeUnitsIndexController extends Controller {
   @tracked municipality = '';
   @tracked province = '';
   @tracked classificationId = '';
+  @tracked selections = '';
```
claire-lovisa, line 30: The name is a bit vague, wouldn't `classificationIds` work ? Or `selectedClassificationIds`, something that shows it's about classifications haha

claire-lovisa (changes requested):
Overall it works well ! Aside from the comments in the PR, I also noticed that when reseting the filters (`Herstel alle filters`) the selected types in the components are not cleared, can you check it out ? Thanks !

app/components/classification-select.hbs
```diff
@@ -6,7 +6,7 @@
     @searchMessage="Typ om te zoeken"
     @searchField="label"
     @options={{this.classifications.value}}
-    @selected={{this.selectedClassification}}
+    @selected={{@selected}}
```
claire-lovisa, line 9: Indeed now that the component isn't used in a mu-search context that change is nice, can you remove the associated getter `get selectedClassification()` from `classification-select.js` as well, as we don't use it anymore ?

app/controllers/administrative-units/index.js
```diff
@@ -79,23 +80,24 @@ export default class AdministrativeUnitsIndexController extends Controller {
...
+  // get isAgbOrApb() {
+  //   return this.isAgb || this.isApb;
+  // }
+  // get isAgb() {
+  //   return this.classificationId === CLASSIFICATION_CODE.AGB;
+  // }
+
+  // get isApb() {
+  //   return this.classificationId === CLASSIFICATION_CODE.APB;
+  // }
```
claire-lovisa, line 92: You can remove them if they're not used

app/controllers/administrative-units/index.js
```diff
@@ -79,23 +80,24 @@ export default class AdministrativeUnitsIndexController extends Controller {
...
+    return this.classificationId.includes(CLASSIFICATION_CODE.WORSHIP_SERVICE);
   }
 
   get isCentralWorshipService() {
-    return (
-      this.classificationId === CLASSIFICATION_CODE.CENTRAL_WORSHIP_SERVICE
+    return this.classificationId.includes(
+      CLASSIFICATION_CODE.CENTRAL_WORSHIP_SERVICE
     );
   }
```
claire-lovisa, line 102: Those aren't used either, no ? As well as `get isWorshipAdministrativeUnit()` ?

app/controllers/administrative-units/index.js
```diff
@@ -27,6 +27,7 @@ export default class AdministrativeUnitsIndexController extends Controller {
   @tracked municipality = '';
   @tracked province = '';
   @tracked classificationId = '';
+  @tracked selectionArray = '';
```
claire-lovisa, line 30: Can you specify that the selection is about classifications ?

claire-lovisa (commented):
Few cleanup comments but it is working well :)

app/components/classification-select.hbs
Dachamelemon: No technically not there only one problem i see with how i did this and that when you reload the page. The field gets cleared while the filter stays active the rest just keeps their value but because I set the value in the index.js and not let it be handled by the component this one does lose it's value.

claire-lovisa: Ah indeed but that is a problem in the multi select component, not in the default select, right ? But good catch, that needs to be fixed before merging

claire-lovisa (approved):

PR #525, opened 2023-10-20
fix dependency worship filter

Dachamelemon:
OP-2771


claire-lovisa (approved):
Thanks for the quick fix !

PR #528, opened 2023-10-24
Fix selected province and municipality in core data related orgs

claire-lovisa:
OP-2788

On first load or after a refresh, `this.selectedProvince` and `this.selectedMunicipality` were not set, causing the conditional selection not to work.


Dachamelemon (approved):
Looks good!

PR #530, opened 2023-10-25
generic orgs

nbittich:
OP-2787 frontend part. this is not complete as I would suggest we should maybe rename everything named with "administrative-unit-***" by something more generic

### DIFF
--- a/app/templates/administrative-units/index.hbs
+++ b/app/templates/administrative-units/index.hbs
@@ -94,7 +94,7 @@
       <t.menu as |menu|>
         <menu.general>
           <PageHeader class="au-o-box">
-            <:title>Bestuurseenheden</:title>
+            <:title>Organisaties</:title>
             <:action>
               <SecuredArea>
                 <:edit>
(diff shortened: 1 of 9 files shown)


app/templates/administrative-units/index.hbs
```diff
@@ -167,13 +167,18 @@
...
-              </LinkTo>
+
+              </AuLink>
+              {{! <LinkTo }}
+              {{!   class="au-c-link" }}
+              {{!--   @model={{adminUnit.id}} --}}
+              {{!--   @route="{{(fn getOrgRoute adminUnit)}}" --}}
+              {{! > }}
+              {{!--   {{adminUnit.name}} --}}
+              {{! </LinkTo> }}
```
claire-lovisa, line 181: This can be removed ?

claire-lovisa (approved):
Working as it is ! But indeed the refactoring cleanup will be much appreciated. We also will be able to merge the `bestuuresenheden` and `organisaties` modules right ?

nbittich:
@claire-lovisa ticket OP-2796 is for the full code refactoring, I'd suggest to wait for ember 4 otherwise it will be a mess 😬 
We probably want to merge module administrative-units & organizations in the code if that's what your meant 🤔

claire-lovisa:
> @claire-lovisa ticket OP-2796 is for the full code refactoring, I'd suggest to wait for ember 4 otherwise it will be a mess 😬 We probably want to merge module administrative-units & organizations in the code if that's what your meant 🤔

Indeed I meant merge as : those two module will be able to become just one !

claire-lovisa (approved):

PR #534, opened 2023-10-31
allow open org in new tab

nbittich:


claire-lovisa (approved):
thanks it's gonna make my life much nicer hahaha

PR #540, opened 2023-11-17
Feat/phonenumber convention sync

Dachamelemon:
OP-1464


app/modifiers/inputmaskTel.js
```diff
@@ -0,0 +1,9 @@
+import { modifier } from 'ember-modifier';
+import Inputmask from 'inputmask';
+
+export default modifier(function inputmaskTel(element) {
+  new Inputmask({
+    regex: '(\\+)\\d{6}\\d*',
```
claire-lovisa, line 6: Shouldn't it be 10 instead of 6 ? For the size of a regular belgian phone number

claire-lovisa (changes requested):
I found an issue I didn't see the first time. If you fill in a number, save the bestuur, then try to re-edit the same page and immediately save, an error will pop up on the phone number saying ` Enkel een plusteken en cijfers zijn toegelaten` even though the number seem to be filled in and formatted correctly

mirdono:
> I found an issue I didn't see the first time. If you fill in a number, save the bestuur, then try to re-edit the same page and immediately save, an error will pop up on the phone number saying ` Enkel een plusteken en cijfers zijn toegelaten` even though the number seem to be filled in and formatted correctly

The same behaviour seems to be present when trying to save an existing adm. unit location that already had a phone number. If you edit all present phone numbers before saving no such error occurs.

app/controllers/administrative-units/administrative-unit/core-data/edit.js
```diff
@@ -150,6 +150,10 @@ export default class AdministrativeUnitsAdministrativeUnitCoreDataEditController
       structuredIdentifierKBO,
     } = this.model;
 
+    yield Promise.all([contact]);
+
+    console.log(contact.telephone);
```
claire-lovisa: This can probably go away ? :)

claire-lovisa (commented):
The bug seems fixed ! :) Just need to wait for the backend PR to be ready to merge both at the same time !

claire-lovisa (approved):

PR #542, opened 2023-11-21
Fix: make names of AGB, APB, and IGS clickable on positions page

mirdono:
OP-2831

Note, only functional change is the removal of the if-then-else clause in the second commit, other changes are purely code formatting.


claire-lovisa (approved):
Perfect ! Can you add a label to the PR as well ? That way it'll be picked up by lerna when creating the release changelog; Thanks !

PR #543, opened 2023-11-22
Tests fix and cleaning

aliokan:
- Fix helper rrn-format test
- Remove empty tests
- Use `this.inputValue` and `strictEqual` in order to avoid deprecated warning


claire-lovisa (approved):
Thanks for the fixes and cleanup :smile:

PR #548, opened 2023-12-11
Avoid scrolling in multiple select on classifcation

lagartoverde:
So this is related to OP-2854, the line avoiding the power select to go full height was defined in appuniversum, as it was set to a max-width of 10rem, I assumed this is something we only want to do in the classification power select, so I restricted it to only be there, but if you think we want this behaviour on all the multiple power select I can change it


claire-lovisa (approved):
Super nice :) I think that we would want this for all multi-select, but for now we only have the classifications one if i remember correctly. But if you could extend it to all multi select just in case, that'd be great :)

PR #549, opened 2023-12-16
Refactor: classification code duplication

mirdono:
An initial refactoring to reduce the code duplication concerning classification codes to reduce the changes needed to add new classification codes in the future.

In the longer run I'd argue we should opt to drop the `CLASSIFICATION_CODE` object in favour of using the `CLASSIFICATION` object everywhere.


claire-lovisa (approved):
Thanks ! And agreed for the drop of `CLASSIFICATION_CODE` at some point :sweat_smile:

PR #545, opened 2023-11-30
Replace `ember-changeset` by a custom form validation solution

aliokan:
**NOTE** The PR is looking for general reviews. It's still a work in progress, but it already gives a good idea of what I'm trying to achieve.

## Context
Currently, the project cannot update Ember and Ember Data due incompatibility between `ember-changeset` library and recent Ember version. 

The goal of this pull request is to define a custom custom form validation solution that replace `ember-changeset`. 

As a first step, only `app/controllers/administrative-units/administrative-unit/core-data/edit.js` and `app/controllers/administrative-units/new.js` forms are converted. The others forms will be managed in follow-up PR. 

## Proposal
Taking inspiration from https://mainmatter.com/blog/2021/12/08/validations-in-ember-apps/
The proposed solution shifts the responsibility of validation into the model itself. 

- models are responsible of there own validation;  
- validations are describe is schema in models; 
- ~~Use `Yup` to describe schema~~. At the end, `Joi` was chosen over `yup` due to limitations encountered with `yup` when validating the address model, specifically related to circular dependencies.
- ~~relationship validation validates recursively relationship~~ The idea of recursively validation in relationships has been removed, because we cannot block a form submission when deep relationships are not valid.

### AbstractValidationModel
A new class `AbstractValidationModel` is introduced. This class incorporates validation capabilities using the `joi` library. Models that require validation should now extend this class and define their validation schema within the `validationSchema` getter.

### Validation Process
Validation can be initiated by calling the `validate()` method on a model instance. The validation result is a boolean indicating whether the model is valid or not. If invalid, error information is accessible through the `error` property.

### Reset Functionality
The AbstractValidationModel also provides a `reset()` method to revert the model to its original state by calling `rollbackAttributes` and set `error` to null. 

 ### Preset of joi schema
A set of `joi` schema functions is available in `validator/schema.js` : 
- validateBelongToRequired(message)
- validateBelongToOptional()
- validateHasManyRequired(message)
- validateHasManyOptional()
- validatePhone(message)
- validateEmail(message)
- validateUrl(message)
- validateRequiredWhenAll(fields, message)
- validateRequiredWhenClassificationId(classificationCodeList, message)

### Usage Example
Models extending `AbstractValidationModel` should implement the `validationSchema` method, specifying the validation rules.

```javascript
export default class OrganizationModel extends AbstractValidationModel {
  @attr name;

  ...

  get validationSchema() {
    return super.validationSchema.append({
      name: Joi.string()
        .required()
[...]

PR #550, opened 2023-12-18
Refactor: classification code selection by role

mirdono:
Looking further at the code duplication for classification codes I experimented a bit in how we can further improve the situation. Since I am not an Ember (or JS) expert this is more of an RFC to see whether this is an appropriate approach or that there are better ways.

The important changes are:
- Added a function `listClassificationCodes` in the model for administrative unit classification codes. This function returns a list of classification codes, and has an argument to choose between worship and non-worship.
- The above function is used in the controller for the `people/index` page to get rid of most of the constants in the `classificationCodes` getter. In that getter I still manually filter 3 organisation types to match the original behaviour.


claire-lovisa:
I do like the approach, in combination with what @aliokan started in here #545 we should end up with way less duplication :balloon:

mirdono:
Superseded by the changes in https://github.com/lblod/frontend-organization-portal/pull/554

PR #553, opened 2023-12-21
Feature: onboarding public ocmw associations

mirdono:
OP-2839, OP-2840, OP-2842, OP-2843, OP-2844, OP-2845

This modifies the frontend to deal with **public** OCMW associations, this covers administrative units of classifications _Welzijnsvereniging_ and _Autonome verzorgingsinstelling_

Backend PR: [https://github.com/lblod/app-organization-portal/pull/367](https://github.com/lblod/app-organization-portal/pull/367)


app/components/change-event-type-select.js
```diff
@@ -70,6 +71,20 @@ export default class ChangeEventTypeSelectComponent extends Component {
     ) {
       types = types.filter((t) => this.isIdInList(t.id, CHANGE_EVENTS_PZ_HPZ));
     }
+    if (
+      classification.id == CLASSIFICATION_CODE.WELZIJNSVERENIGING ||
+      classification.id == CLASSIFICATION_CODE.AUTONOME_VERZORGINGSINSTELLING
+      // ||
```
lagartoverde, line 77: I wouldn't comment out code, if we don't need it just delete it

app/models/administrative-unit-classification-code.js
```diff
@@ -59,10 +59,39 @@ export const CLASSIFICATION = {
     label: 'Hulpverleningszone',
   },
   REPRESENTATIVE_ORGAN: {
-    // fixme this is not an administrative unit
+    // FIXME this is not an administrative unit
     id: '89a00b5a-024f-4630-a722-65a5e68967e5',
     label: 'Representatief orgaan',
   },
+  // Currently not used, split in 5 subtypes below
```
lagartoverde, line 66: If not used then we should delete it

lagartoverde (changes requested):
Some minor comments, I still need to test it

app/components/change-event-type-select.js
mirdono: In most cases I would agree. But this code will be needed when we onboard the private OCMW associations in the near future. You will notice there are similar bits of commented code at other places.

Some background information, initially the goal was to onboard both public and private OCMW associations at the same time. But the latter are _registered organisations_ and not _administrative units_. In its current state the frontend requires more in-depth changes to properly handle non-administrative unit organisations (a refactoring that is also discussed in https://github.com/lblod/frontend-organization-portal/pull/530). Therefore, we decided to onboard OCMW associations in two phases: now the public ones, and the private ones after refactoring the frontend. At the time of this decision I had already added some code that deals with private OCMW associations. To me it made more sense to temporarily comment out this code, compared to removing it only to re-add it latter on.

lagartoverde (approved):

PR #554, opened 2023-12-22
POC custom form validation with `joi`

aliokan:
## Context
Currently, the project cannot update Ember and Ember Data due incompatibility between `ember-changeset` library and recent Ember version. 

The goal of this pull request is to define a new custom form validation solution that replace `ember-changeset`. 

As a first step, only the following forms are converted: 
- `/bestuurseenheden/nieuwe-bestuurseenheid`
- `/bestuurseenheden/_id_/kerngegevens/edit`
- `/bestuurseenheden/_id_/vestigingen/nieuwe-vestiging`
- `/bestuurseenheden/_id_/vestigingen/_id_/edit`

The others forms will be managed in follow-up PR. 

## Proposal

The proposed solution shifts the responsibility of validation into the model itself. 

- models are responsible of there own validation;  
- validations are describe in `joi` schema in models; 

### AbstractValidationModel
A new class `AbstractValidationModel` is introduced. Models that require validation should now extend this class and define their validation schema within the `validationSchema` getter.

### Validation Process
Validation can be initiated by calling the `validate()` method on a model instance. The validation result is a boolean indicating whether the model is valid or not. If invalid, error information is accessible through the `error` property.

### Reset Functionality
The AbstractValidationModel also provides a `reset()` method to revert the model to its original state by calling `rollbackAttributes` and set `error` to null. 

 ### Preset of joi schema
A set of `joi` schema functions is available in `validator/schema.js` : 
- validateBelongToRequired(message)
- validateBelongToOptional()
- validateHasManyRequired(message)
- validateHasManyOptional()
- validatePhone(message)
- validateEmail(message)
- validateUrl(message)
- validateRequiredWhenAll(fields, message)
- validateRequiredWhenClassificationId(classificationCodeList, message)

### Usage Example
Models extending `AbstractValidationModel` should implement the `validationSchema` method, specifying the validation rules.

```javascript
export default class ExampleModel extends AbstractValidationModel {
  @attr name;

  get validationSchema() {
    return Joi.object({
      name: Joi.string()
        .required()
        .messages({ 'any.required': 'Vul de naam in' }),
    });
  }
}
```

Sub-class can extends parent schema to add more rules: 

```javascript
export default class WorshipAdministrativeUnitModel extends AdministrativeUnitModel {
  @belongsTo('recognized-worship-type', {
    inverse: null,
  })
  recognizedWorshipType;

  @hasMany('minister-position', {
    inverse: 'worshipService',
  })
  ministerPositions;

  @hasMany('local-involvement', {
    inverse: 'worshipAdministrativeUnit',
  })
  involvements;

  get validationSchema() {
    return super.validationSchema.append({
      recognizedWorshipType: validateRequiredWhenClassificationId(
        WorshipServiceCodeList,
        'Selecteer een optie'
      ),
      ministerPositions: validateHasManyOptional(),
      involvements: validateHasManyOptional(),
    });
  }
}
```

## TODO
- [x] Remove `console.log`


app/models/administrative-unit.js
```diff
@@ -31,4 +51,103 @@ export default class AdministrativeUnitModel extends OrganizationModel {
...
+      ),
+      isSubOrganizationOf: validateRequiredWhenClassificationId(
+        [
+          ...AgbCodeList,
+          ...ApbCodeList,
+          ...IGSCodeList,
+          ...PoliceZoneCodeList,
+          ...AssistanceZoneCodeList,
+          ...WorshipServiceCodeList,
+          ...CentralWorshipServiceCodeList,
```
mirdono, line 88: In the current changeset-based validation we do not enforce this relation for worship units. I am not sure based on what you added it here and whether it might break something somewhere.

mirdono:
Imo this is an improvement over the existing changeset-based validations and I suggest we move forward with applying this to all forms in the OP app, unless there are other objections.

mirdono (commented):
Imo this is an improvement over the existing changeset-based validations and I suggest we move forward with applying this to all forms in the OP app.

mirdono (approved):

PR #555, opened 2023-12-22
Feat/integration wegwijs api

Dachamelemon:
Frontend KBO tab added with temporary static data until backend is ready.

the new class added in _shame.scss should be made option in the Tab component of appuniversum followed up by @oscartison 

Feature Flag:
add "?feature-kbo-data-tab=true" to URI to see feature.


app/templates/administrative-units/administrative-unit/core-data/index.hbs
mirdono, line 1: When the KBO tab is selected the edit button remains visible and, when clicked, will redirect the user to the already existing edit page (i.e. the page to edit the ABB data). This is imo confusing from a UX perspective, and is not conform with the proposed design for this page.

mirdono (changes requested):

mirdono (approved):

aliokan (changes requested):
**Blocker** This PR cannot be merged in state. As discuss during standup, the current modifications should be hidden behind feature flag. ;)

aliokan (approved):
The feature flag integration looks good! :+1:

PR #557, opened 2024-01-09
Add feature flag service and helper

aliokan:
Import Feature flag service and helper from [Lokaal Beslist](https://github.com/lblod/frontend-burgernabije-besluitendatabank/tree/master#feature-flags). 

## How to use it? 

1. Define a new flag in the [config/environment.js](https://github.com/lblod/frontend-organization-portal/blob/master/config/environment.js)
```javascript
// in config/environment.js
let ENV = {
  // Other configuration settings...
  features: {
    "new-feature": true, // Enable the 'new-feature' by default
    "beta-feature": false, // Disable the 'beta-feature' by default
  },
};
```
2. Wrap your modifications behind the flag : 

With the service in router/controller/components
```javascript
import { inject as service } from "@ember/service";

export default class ExampleComponent extends Component {
  @service features;

  doSomething() {
    if (this.features.isEnabled("new-feature")) {
      // Implement the logic for the new feature
      console.log("New feature is enabled!");
    } else {
      // Implement the logic for the default behavior without the new feature
      console.log("New feature is disabled!");
    }
  }
}
```
With the helper in template
```handlebars
{{#if (is-feature-enabled "new-feature")}}
  <p>New feature is enabled!</p>
{{else}}
  <p>New feature is disabled!</p>
{{/if}}
```

## How to test? 
The default flag value in configuration can be manually overridden by adding a query parameter to the URL:

- `?feature-new-feature=false` to disable the 'new-feature'
- `?feature-beta-feature=true` to enable the 'beta-feature'

The overriding will be saved in a cookie, so it will persist across page reloads. The cookie can be cleared by adding `?clear-feature-overrides=true` to the URL or manually remove the cookie in developer tools.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,3 +57,62 @@ Specify what it takes to deploy your app.
 ## Docker
 
 Build the image: docker build -t lblod/frontend-organization-portal -f Dockerfile.fastboot .
+
+## Reference
+
+### Feature flags
+
+Feature flags are used to enable/disable features in the application. They are defined in [config/environment.js](config/environment.js).
+
+```javascript
+// in config/environment.js
+let ENV = {
+  // Other configuration settings...
+  features: {
+    "new-feature": true, // Enable the 'new-feature' by default
+    "beta-feature": false, // Disable the 'beta-feature' by default
+  },
+};
+```
+
+The configuration can be manually overridden by adding a query parameter to the URL:
+
+- `?feature-new-feature=false` to disable the 'new-feature'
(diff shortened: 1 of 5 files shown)


mirdono (commented):
Works as advertised and looks good to me. One minor remark, could you also add the usage instructions to the README, as was done in lokaalbeslist?

aliokan:
@mirdono Doc added

mirdono (approved):

PR #565, opened 2024-01-18
Migrate `administrative-unit.related-organizations.edit` to the new form validation

aliokan:
Related to OP-2914

Migrate `administrative-units.administrative-unit.related-organizations.edit` form to the new `administrative-unit` form validation defined in https://github.com/lblod/frontend-organization-portal/pull/554


app/controllers/administrative-units/administrative-unit/related-organizations/edit.js
```diff
@@ -14,100 +14,13 @@ export default class AdministrativeUnitsAdministrativeUnitRelatedOrganizationsEd
...
-  }
-
-  get isCentralWorshipService() {
-    return (
-      this.model.administrativeUnit.classification?.get('id') ===
-      CLASSIFICATION_CODE.CENTRAL_WORSHIP_SERVICE
-    );
-  }
-
   get hasCentralWorshipService() {
```
mirdono, line 24: Imo this getter should also move to the model. It is duplicated in the [controller](https://github.com/lblod/frontend-organization-portal/blob/a2dbad5012c5d3aa0c5e2dfaa261588a3821b91a/app/controllers/administrative-units/new.js#L30) for new administrative units. (A similar argument can be made for moving (some) of the other remaining getters to the classification constants, but that is out of scope for this PR.)

mirdono (commented):
Looks good to me, aside from the minor comment concerning the `hasCentralWorshipService` getter.

app/controllers/administrative-units/administrative-unit/related-organizations/edit.js
aliokan: I hesitated, then I follow the same logic as the POC. But I agree it should be migrated to. I propose to migrate all remaining duplicate getters in controllers in a follow-up PR, between the form validation migration and the Ember update.
@mirdono What do you think?

mirdono: Dealing with the remaining duplicate getters, including the one I pointed out here, in a separate PR is fine for me. So no further objections for this PR.

mirdono (approved):

PR #568, opened 2024-01-24
Feat/site names

lagartoverde:
OP-2905


app/controllers/administrative-units/administrative-unit/sites/new.js
```diff
@@ -20,6 +21,14 @@ export default class AdministrativeUnitsAdministrativeUnitSitesNewController ext
     );
   }
 
+  get isAndereVestigen() {
```
mirdono, line 24: This getter should be placed in the `site` model such that it does not need to be duplicated here and in the edit route.  (This is a change coming with new form validation framework.)
Concerning the function name, I am inclined to use English names to avoid (imo) unnecessary hurdles for non Dutch speakers to read/understand the code, so in this case something like `isOtherSite` seems appropriate. If you do want to keep the Dutch name it should be `isAndereVestiging` as "Vestig**en**" is not correct in this context.

app/models/site.js
```diff
@@ -23,11 +23,14 @@ export default class SiteModel extends AbstractValidationModel {
...
   siteType;
 
+  @attr siteTypeName;
+
   get validationSchema() {
     return Joi.object({
       address: validateBelongsToOptional(),
       contacts: validateHasManyOptional(),
       siteType: validateBelongsToRequired('Voeg een type vestiging toe'),
+      siteTypeName: Joi.string().allow('').optional(),
```
mirdono, line 33: **Nitpick:**  As I understand it specifying both `allow('')` and `optional()` is redundant here. Specifying `allow('')` (or `empty('') as done in [another model](https://github.com/lblod/frontend-organization-portal/blob/2698fe92c0a3e9584a146467e2defd6cc00f0145/app/models/organization.js#L101)) should be sufficient.

app/templates/administrative-units/administrative-unit/sites/site/index.hbs
```diff
@@ -31,6 +31,12 @@
               <:label>Type vestiging</:label>
               <:content>{{@model.site.siteType.label}}</:content>
             </Item>
+            {{#if this.isAndereVestigen}}
+              <Item>
+                <:label>Name</:label>
+                <:content>{{@model.site.siteTypeName}}</:content>
+              </Item>
+            {{/if}}
```
mirdono, line 39: **Nitpick:** do we want to display this label even if there is no value for name? I think that optional fields without a value are generally not displayed in this frontend.

app/controllers/administrative-units/administrative-unit/sites/new.js
lagartoverde: Done :)

app/models/site.js
lagartoverde: yeah, had a small fight with Joi for this PR xd, and this is what it worked, but you are right, empty("") works also fine

mirdono (approved):

PR #570, opened 2024-01-30
Refactor: migrate governing body form validation to Joi

mirdono:
OP-2910

I attempted to replicate the behaviour of the previous validation as close as possible, although this leads to some (imo) unintuitive/unexpeced aspects:
- Whether or not the periods of governing bodies overlap is **not** symmetrical. For example, it is allowed to specify a start and end date such that the period of the governing being edited is equal or contained within the period of another governing body. The inverse, specifying a start and end date such that they would contain the period of another governing body, is not allowed.
-  The error message for overlap "Geen overlap" is displayed on both the start and end date fields, irrelevant of which of these caused the overlap.
- The error messages concerning an incorrect order of start and end date (e.g. the start date is after the end date) are displayed on both form fields, irrelevant whether only one of them was changed.

Other remarks:
- The relation between an administrative unit and its **timed** governing bodies, i.e. those the user edits, is an indirect one. An administrative unit has zero or more classes of **untimed** governing bodies. Each untimed body, in turn, has zero or more time specialisations. These time specialisations do **not** directly refer back to their corresponding administrative unit, i.e. their `administrativeUnit` relation is not set. Consequently, retrieving the appropriate governing bodies to check for overlap turned out to be more complicated than anticipated and results in re-retrieving data that was already in the route's model (cf. the functions at the end of the governing body model). Any suggestions for improvements there would be welcome.


tests/unit/models/governing-body-test.js
```diff
@@ -0,0 +1,828 @@
...
+      const model = this.store().createRecord('governing-body');
+
+      const isValid = await model.validate();
+
+      assert.false(isValid);
+      assert.strictEqual(Object.keys(model.error).length, 2);
+      assert.strictEqual(model.error.startDate.message, 'Vul de startdatum in');
+      assert.propContains(model.error, {
+        endDate: { message: 'Vul de einddatum in' },
+      });
```
aliokan, line 25: **nitpick:** What about moving `startDate.message` error in the `propContains` object? ;)

tests/unit/models/governing-body-test.js
```diff
@@ -0,0 +1,828 @@
...
+      assert.notPropContains(model.error, {
+        startDate: {
+          message: 'Kies een startdatum die vóór de einddatum plaatsvindt',
+        },
+      });
+      assert.notPropContains(model.error, {
+        endDate: {
+          message: 'Kies een einddatum die na de startdatum plaatsvindt',
+        },
+      });
```
aliokan, line 49: **note:** if the `strictEqual` and `propContains` assert are valide , then `notPropContains` assert will all the time be true and are redundant.

tests/unit/models/governing-body-test.js
```diff
@@ -0,0 +1,828 @@
...
+      assert.strictEqual(Object.keys(model.error).length, 2);
+      assert.propContains(model.error, {
+        startDate: {
+          message: 'Kies een startdatum die vóór de einddatum plaatsvindt',
+        },
+        endDate: {
+          message: 'Kies een einddatum die na de startdatum plaatsvindt',
+        },
+      });
+    });
```
aliokan, line 181: It's a duplicate of the previous one ;)

tests/unit/models/governing-body-test.js
```diff
@@ -0,0 +1,828 @@
...
+      const otherBodiesStub = sinon.stub(model, 'getOtherTimedGoverningBodies');
+      otherBodiesStub.resolves([otherModel]);
+
+      const isValid = await model.validate();
+
+      assert.true(isValid);
+      assert.notOk(model.error, 'model.error should be undefined');
+
+      otherBodiesStub.restore();
+    });
```
aliokan, line 281: Really similar to the previous one

tests/unit/models/governing-body-test.js
```diff
@@ -0,0 +1,828 @@
...
+        },
+        endDate: {
+          message: 'Geen overlap',
+        },
+      });
+
+      otherBodiesStub.restore();
+    });
+
+    test("it returns no error when the model's period is strictly within the period of another governing body", async function (assert) {
```
aliokan, line 314: **question:** Is it the expected behavior? It isn't overlapping?

tests/unit/models/governing-body-test.js
```diff
@@ -0,0 +1,828 @@
...
+      const otherBodiesStub = sinon.stub(model, 'getOtherTimedGoverningBodies');
+      otherBodiesStub.resolves([otherModel]);
+
+      const isValid = await model.validate();
+
+      assert.true(isValid);
+      assert.notOk(model.error, 'model.error should not be undefined');
+
+      otherBodiesStub.restore();
+    });
```
aliokan, line 512: **thought:** All these tests looks really close, isn't possible to group them in a for loop?

tests/unit/models/governing-body-test.js
```diff
@@ -0,0 +1,828 @@
...
+      const changedStub = sinon.stub(model, 'changedAttributes');
+      changedStub.returns({});
+
+      const isValid = await model.validate();
+
+      assert.true(isValid);
+      assert.notOk(model.error, 'model.error should be undefined');
+
+      changedStub.restore();
+    });
```
aliokan, line 546: These tests are similar, only one can be keep and rename is something more generic like: "it should not check date overlap unless it was changed"

tests/unit/models/governing-body-test.js
```diff
@@ -0,0 +1,828 @@
...
+      const governingBodies = await timedBody.getOtherTimedGoverningBodies();
+      assert.strictEqual(governingBodies.length, 0);
+
+      queryStub.restore();
+    });
+
+    test('it returns an array containing the other timed governing bodies when there is one of the same classification', async function (assert) {
+      const administrativeUnit = this.store().createRecord(
+        'administrative-unit',
+        { id: '1234' }
```
aliokan, line 604: **note:** I notice that you are adding `id` in almost all `createRecord` but they are not required, unless you need to use them in the test they can be removed.

aliokan (changes requested):
@mirdono I left some comments, nothing really serious. :ping_pong:

tests/unit/models/governing-body-test.js
mirdono: The set `id` is used in `#getAdministrativeUnitId` which is called in `#getUntimedGoverningBodies` which  in turn is called by the function being tested `getOtherGoverningBodies`. Not specifying an `id` causes this functionality, and thereby also the tests, to fail in an unintended way.

tests/unit/models/governing-body-test.js
mirdono: Not sure whether it is the expected behaviour, but that is how it worked before migrating to the Joi validation. So I kept it for this refactoring, although I agree that this situation is overlapping. It is probably a good idea to have a follow up to look at what "overlap" precisely means in the context of governing bodies and change our implementation accordingly.

tests/unit/models/governing-body-test.js
mirdono: I agree that these two tests are currently the same, but I would opt to keep them both. I think it would make more sense to show an error only for a field if it was changed, instead of just always showing both error messages as is the current behaviour (copied from how it was before.) Imo having both tests already there can be beneficial when making such changes in the future, although this is clearly debatable.

tests/unit/models/governing-body-test.js
mirdono: Huh, no idea why I did it differently in those two cases.

app/models/governing-body.js
```diff
@@ -35,15 +43,138 @@ export default class GoverningBodyModel extends Model {
...
+    let administrativeUnitId = await this.#getAdministrativeUnitId();
+    let untimedGoverningBodies = await this.store.query('governing-body', {
+      filter: {
+        'administrative-unit': {
+          id: administrativeUnitId,
+        },
+      },
+      include: 'has-time-specializations,classification',
+    });
+    return untimedGoverningBodies;
```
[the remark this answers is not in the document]
mirdono: Was not aware that the filter would be ignored, an extra check seems appropriate in that case.

app/models/governing-body.js
```diff
@@ -35,15 +43,138 @@ export default class GoverningBodyModel extends Model {
...
+              // pre-Joi validation resulted in error messages for both date
+              // form fields, hence the check of the other body's end date here
+              // as well.
+              if (
+                inPeriod(body.startDate, this.startDate, this.endDate) ||
+                inPeriod(body.endDate, this.startDate, this.endDate)
+              ) {
+                return helpers.message('Geen overlap');
+              }
+            }
```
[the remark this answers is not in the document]
mirdono: Agreed, the validations for start and end date were initially different from each other. But to replicate the original behaviour I changed it, resulting in the duplicated code.

tests/unit/models/governing-body-test.js
mirdono: I agree that the tests are very similar, but due to the asymmetric definition of overlap that is used I prefer to have tests to explicitly cover all cases.

tests/unit/models/governing-body-test.js
mirdono: I merged these together in single test with loop.

tests/unit/models/governing-body-test.js
aliokan: Can you add a Technical story in the backlog to not forget?

aliokan (changes requested):
@mirdono  I still have one little blocker :ping_pong:

app/models/governing-body.js
```diff
@@ -35,15 +43,135 @@ export default class GoverningBodyModel extends Model {
...
+        include: 'has-time-specializations,classification',
+      });
+      return untimedGoverningBodies;
+    }
+  }
+
+  async getOtherTimedGoverningBodies() {
+    let untimedGoverningBodies = (
+      await this.#getUntimedGoverningBodies()
+    ).toArray();
```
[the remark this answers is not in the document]
mirdono: I made the implementation a bit more robust with respect to missing values/relations, and added the corresponding tests.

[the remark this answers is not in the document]
aliokan: Thank you! :pray:

aliokan (approved):
Everything looks good, thanks for the improvements! :rocket:

PR #572, opened 2024-02-02
Fix address and contact saving

aliokan:
Related to OP-2974

Replace remaining `isDirty` provided by `ember-changeset` by `hasDirtyAttributes`.


mirdono (approved):
Looks good to me, taking into account your remark in #571 concerning the remaining `isDirty`.

PR #562, opened 2024-01-16
Remove personal-information and minister disabled forms

aliokan:
Related to OP-2934

Remove forms :
- people.person.positions.minister.edit
- people.person.personal-information.edit
- administrative-units.administrative-unit.ministers.new
- people.new-position (keep route and message, but remove form related content)

Remove `ember-changeset` rules 
- validations/address.js
- validations/contact-point.js
- validations/minister.js
- validations/secondary-contact-point.js
- validations/site.js

Remove components: 
- contact-details
- propagate-contact-details-update

PR #573, opened 2024-02-02
Fix: unable to leave page 'personen/nieuw-positie'

aliokan:
Related to OP-2984

## Context 

Error was due to missing `this.router` in controller removed during https://github.com/lblod/frontend-organization-portal/pull/562

## Proposal

I replace `AuButton` by `AuLink` in order to remove the controller.


mirdono (approved):
From what I tested this fixes the bug.

PR #580, opened 2024-02-19
Fix: added missing classification codes

mirdono:
OP-3032

In #553 and #558 the `selectByRole` function used when searching on name, was not (correctly) modified. This resulted in a bug where organisations of some of the newly introduced classifications were not included in the search results.


aliokan (approved):
Thank you! :pray:

PR #581, opened 2024-02-27
Fix: show related organisations in edit core data form for PEVAs

mirdono:
Related to OP-3047

The reported bug also revealed that the edit core data form contains insufficient fields in case the organisation being edited is a PEVA municipality or PEVA province. More specifically, the form lacks the related organizations part in such case. This PR corrects this particular form so that it shows the correct fields for both types of PEVAs. 

Note: this PR is **not** intended to fix the reported bug itself, i.e. the inability to edit core data of PEVAs that do not have a founding organisation.


aliokan (approved):
Look good! :+1:

PR #578, opened 2024-02-07
DRAFT: integrated first wegwijs data in frontend

Dachamelemon:
This is a draft.
Linked to backend branch: https://github.com/lblod/app-organization-portal/tree/feat/wegwijs-service-integration

PR #582, opened 2024-02-28
Feat: add legal name for organisations

mirdono:
OP-2998 and OP-2999

# Summary
- Added a mandatory `legalName` attribute to the `organization` model, `name` attribute is no longer mandatory
- Replaced displaying of the `name` attribute by a function `abbName` that returns the `legalName` and falls back to `name` if that is not available
- Extended the (edit) core data forms with fields for `legalName`
- After creation, an organisation's legal name can only be edited for some classifications (see table in OP-2999, note during standup of 29/02 we decided that for municipalities and provinces the legal name should also be not editable either)
- The `name` attribute can no longer be directly edited by users
- When an organisation's legal name is edited, the value of both `legalName` and `name` is updated
- Extended the name search to include the name as well as legal name (requires a reindex to work)

# Notes
- The changes to `core-data/index.hbs` are mostly automatic formatting. The relevant changes in that file are the changes from `name` to `abbName` to print an organisation's name and the addition of the items for displaying the legal name.

# TODO
- [ ] Use registered KBO name as fallback value for ABB name, depends on #578

### DIFF
--- a/app/models/organization.js
+++ b/app/models/organization.js
@@ -10,6 +10,7 @@ import {
 
 export default class OrganizationModel extends AgentModel {
   @attr name;
+  @attr legalName;
   @attr alternativeName;
   @attr('date') expectedEndDate;
   @attr purpose;
@@ -134,8 +135,9 @@ export default class OrganizationModel extends AgentModel {
 
   get validationSchema() {
     return super.validationSchema.append({
-      name: Joi.string().empty('').required().messages({
-        'any.required': 'Vul de naam in',
+      name: validateStringOptional(),
+      legalName: Joi.string().empty('').required().messages({
+        'any.required': 'Vul de juridische naam in',
       }),
       alternativeName: validateStringOptional(),
       expectedEndDate: Joi.date().allow(null),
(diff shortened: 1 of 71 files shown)


app/models/organization.js
```diff
@@ -134,8 +135,9 @@ export default class OrganizationModel extends AgentModel {
 
   get validationSchema() {
     return super.validationSchema.append({
-      name: Joi.string().empty('').required().messages({
-        'any.required': 'Vul de naam in',
+      name: Joi.string().empty(''),
```
aliokan, line 138: **Suggestion (blocker):** In some case an input field value is null, null will be reject because it's not a string, we have `validateStringOptional` to avoid this kind of type error : https://github.com/lblod/frontend-organization-portal/blob/development/app/validators/schema.js#L96-L98

app/templates/administrative-units/index.hbs
```diff
@@ -168,7 +168,11 @@
           <c.body as |adminUnit|>
             <td>
               <AuLink @route={{adminUnit.route}} @model={{adminUnit.id}}>
-                {{adminUnit.name}}
+                {{#if adminUnit.legal_name}}
+                  {{adminUnit.legal_name}}
+                {{else}}
+                  {{adminUnit.name}}
+                {{/if}}
```
aliokan, line 175: **Question:** Why not use `abbName` here?

aliokan (changes requested):
@mirdono I left 2 small remarks :ping_pong:

app/templates/administrative-units/index.hbs
aliokan: make sense!

aliokan (approved):
I quickly review the template part, I see nothing wrong. :+1:

PR #583, opened 2024-02-29
Fix: relax requirement of founding organization for newly onboarded unit types

mirdono:
OP-3051
Related to the bugs reported in OP-3046 and OP-3047

While a founding organisation should be mandatory for OCMW associations and PEVAs the available business data is incomplete. Strictly enforcing this rule would thus mean that units lacking a founder at the time of onboarding cannot be edited later on as their validation will fail. This commit relaxes this rule such that a founding organisation is only required when creating a new OCMW association or PEVA via the frontend.

### DIFF
--- a/app/models/abstract-validation-model.js
+++ b/app/models/abstract-validation-model.js
@@ -25,12 +25,10 @@ export default class AbstractValidationModel extends Model {
 
   /**
    * Validate the model using the validation schema.
-   * @param {boolean} relaxMandatoryFoundingOrganization - if true, the
-   * requirement to specify a founding organisation is not enforced for OCMW
-   * associations and PEVAs
+   * @param {object} [options] - The options for validation.
    * @returns {Promise<boolean>} - Whether the model is valid.
    */
-  async validate(relaxMandatoryFoundingOrganization) {
+  async validate(options = {}) {
     this._validationError = undefined;
     const { attributes, relationships } = this.#serializeAll();
     const value = { ...attributes, ...relationships };
@@ -40,7 +38,7 @@ export default class AbstractValidationModel extends Model {
         abortEarly: false,
         context: {
           changedAttributes: this.changedAttributes(),
-          relaxMandatoryFoundingOrganization,
+          ...options,
         },
       });
     } catch (error) {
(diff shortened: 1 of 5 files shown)


app/models/abstract-validation-model.js
```diff
@@ -25,9 +25,12 @@ export default class AbstractValidationModel extends Model {
 
   /**
    * Validate the model using the validation schema.
+   * @param {boolean} relaxMandatoryFoundingOrganization - if true, the
+   * requirement to specify a founding organisation is not enforced for OCMW
+   * associations and PEVAs
    * @returns {Promise<boolean>} - Whether the model is valid.
    */
-  async validate() {
+  async validate(relaxMandatoryFoundingOrganization) {
```
aliokan, line 33: **Suggestion (blocker):** `relaxMandatoryFoundingOrganization` is really specific. I'm pretty certain that we will need in the future to add more logic like that. What about passing an object instead of a boolean? It will increase the flexibility, and avoid a long list of arguments in the future :wink: 

**Proposal**
```javascript
  /**
   * Validate the model using the validation schema.
   * @param {object} [options] - The options for validation.
   * @returns {Promise<boolean>} - Whether the model is valid.
   */
  async validate( options = {} ) {
    this._validationError = undefined;
    const { attributes, relationships } = this.#serializeAll();
    const value = { ...attributes, ...relationships };

    try {
      await this.validationSchema.validateAsync(value, {
        abortEarly: false,
        context: {
          changedAttributes: this.changedAttributes(),
          ...options,
        },
      });
```
The validate method should be called `administrativeUnit.validate({relaxMandatoryFoundingOrganization: true})`

app/models/administrative-unit.js
```diff
@@ -76,15 +76,25 @@ export default class AdministrativeUnitModel extends OrganizationModel {
...
-        [
-          ...AgbCodeList,
-          ...ApbCodeList,
-          ...OcmwAssociationCodeList,
-          ...PevaMunicipalityCodeList,
-          ...PevaProvinceCodeList,
-        ],
-        REQUIRED_MESSAGE
+      wasFoundedByOrganizations: Joi.when(
+        Joi.ref('$relaxMandatoryFoundingOrganization'),
```
aliokan, line 80: **Nitpick:** What about adding a comment to explain were `$relaxMandatoryFoundingOrganization` is coming from?

aliokan (changes requested):
@mirdono I add a suggestion :ping_pong:

PR #587, opened 2024-03-11
Fix sharepoint and kbo reset in administrative-unit edit form

aliokan:
Related to OP-3048

KBO and Sharepoint were not correctly reset in administrative-unit edit form controller.


mirdono (approved):
That seems to fix the bug

PR #589, opened 2024-03-12
Update ember-data version to 4.12.0

aliokan:
Related to OP-3065

KBO-number shouldn't be checked if the value hasn't change. https://github.com/lblod/frontend-organization-portal/blob/development/app/models/identifier.js#L41-L44
The previous ember update introduce a bug in `changedAttributes` result after saving https://github.com/emberjs/data/issues/8329

Bump to ember-data 4.12.5 fixes the problem.


mirdono (approved):
It fixes the reported bug as far as I tested.

PR #590, opened 2024-03-13
Fix/added kbo tabs representative bodies

Dachamelemon:
CONTEXT:

during testing a bug a reported that representative bodies had no working data tab for kbo.
Data tab is working and data is also shown if there is any

Merged into: https://github.com/lblod/frontend-organization-portal/pull/578


Dachamelemon:
Merged into: https://github.com/lblod/frontend-organization-portal/pull/578

PR #592, opened 2024-03-18
Fix: names help text

mirdono:
OP-3105


aliokan (approved):
I guess, this is correct ^^

PR #594, opened 2024-03-22
Fix: display "Naam" as label for the name field for some types of sites

mirdono:
Sites of type "Andere vestiging" (Other site) can have an optional name. The label shown for this field was the English "Name" instead of the Dutch "Naam".


aliokan (approved):
Good catch! :monocle_face:

PR #596, opened 2024-03-26
Fix: explicitly copy legal and alternative names when creating new units

mirdono:
OP-3129

The juridical name and alternative names were not explicitly copied to the `newAdministrativeUnit` object that would eventually be saved. Consequently, these names were lost upon saving a worship administrative unit. For non-worship administrative units this did not occur because the created record was reused.


aliokan (approved):
Approved! :+1:

PR #598, opened 2024-04-02
Feat/integrated first wegwijs data in frontend

Dachamelemon:
Due to a very dumb oversight of myself and accidentally merging before review I had to revert everything and try to fix the branch again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,41 @@
 
+
+
+## v1.23.2 (2024-04-11)
+
+#### :bug: Bug Fix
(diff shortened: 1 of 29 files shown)


aliokan (approved):
@Dachamelemon I applied the following changes: 
- Extract the kbo tab template in a component in order to avoid the code duplication
- Fix the date in `Laatst Gewijzigd op` and `Begindatum`
- Restore the call to `sync-ovo-number` service when `kbo-data-tab` flag is disabled
- Simplify the way of loading kboContact in route classes
- Revert the unwanted changes in template certainly due to a merge conflict
- Remove the code duplication in templates

PR #599, opened 2024-04-05
Fix: Validate Sharepoint identifier only when modified

mirdono:
OP-3134

### DIFF
--- a/tests/unit/models/identifier-test.js
+++ b/tests/unit/models/identifier-test.js
@@ -191,7 +191,7 @@ module('Unit | Model | identifier', function (hooks) {
         const structuredIdentifier = this.store().createRecord(
           'structured-identifier',
           {
-            localId: '123',
+            localId: '123abc',
           }
         );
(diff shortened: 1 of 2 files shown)


tests/unit/models/identifier-test.js
```diff
@@ -187,6 +187,28 @@ module('Unit | Model | identifier', function (hooks) {
         assert.true(isValid);
       });
 
+      test('it does not validate when localId was not changed', async function (assert) {
+        const structuredIdentifier = this.store().createRecord(
+          'structured-identifier',
+          {
+            localId: '123',
```
aliokan, line 194: **Issue: ** I suppose the test will pass even with a `changedAttributes`. Can you add some letters in the `localId` value? ;)

aliokan (changes requested):
@mirdono Just one comment regarding the test ;)

aliokan (approved):
To infinity and beyond! :rocket:

PR #600, opened 2024-04-09
Do not trigger validation on empty SharePoint identifier

mirdono:
OP-3147

Note: the second commit is a quick fix for another bug encountered in the process of solving the original one. I think a cleaner solution for that requires more in-depth changes.

### DIFF
--- a/app/controllers/administrative-units/administrative-unit/core-data/edit.js
+++ b/app/controllers/administrative-units/administrative-unit/core-data/edit.js
@@ -195,9 +195,13 @@ export default class AdministrativeUnitsAdministrativeUnitCoreDataEditController
       yield structuredIdentifierKBO.save();
       yield identifierKBO.save();
 
-      structuredIdentifierSharepoint = setEmptyStringsToNull(
-        structuredIdentifierSharepoint
-      );
+      // FIXME: If uncommented existing SharePoint identifier is not removed
+      // when user removes the value in the form. Commenting it a quick, dirty
+      // fix because it overwrites the previous value with an empty string
+      // instead of leaving it untouched.
+      // structuredIdentifierSharepoint = setEmptyStringsToNull(
+      //   structuredIdentifierSharepoint,
+      // );
       yield structuredIdentifierSharepoint.save();
       yield identifierSharepoint.save();
 
--- a/app/models/identifier.js
+++ b/app/models/identifier.js
@@ -67,7 +67,7 @@ export default class IdentifierModel extends AbstractValidationModel {
             const changedAttributes = (
               await this.structuredIdentifier
             ).changedAttributes();
-            if (changedAttributes?.localId && !localId.match(/^\d+$/)) {
+            if (changedAttributes?.localId && !localId.match(/^\d*$/)) {
               return helpers.message(
                 'De SharePoint identificator mag enkel cijfers bevatten'
               );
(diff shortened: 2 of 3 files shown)


app/models/identifier.js
```diff
@@ -67,7 +67,11 @@ export default class IdentifierModel extends AbstractValidationModel {
             const changedAttributes = (
               await this.structuredIdentifier
             ).changedAttributes();
-            if (changedAttributes?.localId && !localId.match(/^\d+$/)) {
+            if (
+              localId &&
+              changedAttributes?.localId &&
+              !localId.match(/^\d+$/)
+            ) {
```
aliokan, line 74: **Suggestions:** Use the optional chaining (?.) will have the same effect and be more consistent with `changedAttributes?.localId`. 

```suggestion
            if (changedAttributes?.localId && !localId?.match(/^\d+$/)) {
```

app/controllers/administrative-units/administrative-unit/core-data/edit.js
```diff
@@ -195,9 +195,13 @@ export default class AdministrativeUnitsAdministrativeUnitCoreDataEditController
       yield structuredIdentifierKBO.save();
       yield identifierKBO.save();
 
-      structuredIdentifierSharepoint = setEmptyStringsToNull(
-        structuredIdentifierSharepoint
-      );
+      // FIXME: If uncommented existing SharePoint identifier is not removed
```
aliokan, line 198: **Question:** Any error message in the console when you try to save with null value?

aliokan (commented):
@mirdono I left comment

app/models/identifier.js
mirdono: In this case it will not have the same effect. 
- For an empty string `?.` will succeed, after which the `match` fails due to the `+` in the regex. The negation will cause the result to be truthy. This turns the whole condition truthy, which triggers a validation error.
- For `null` the behaviour also changes, although I am not 100% sure on why. I _think_ the `changedAttributes?.localId` succeeds when `localId` is `null` (but not for `undefined`?) and the match statement succeeds in both cases as `?.` results in `undefined` which is in turn negated to truthy.

In trying to to understand the above behaviour a more logical solution popped in my head. Namely, changing the `+` in the match regex to a `*` so that 0 or more numbers are valid instead of 1 or more.
```suggestion
            if (changedAttributes?.localId && !localId.match(/^\d*$/)) {
```
My test for `null` as value for `localId` still fails with this change, but that does not seem to a scenario I can cause via the frontend.

app/controllers/administrative-units/administrative-unit/core-data/edit.js
mirdono: Nope. As mentioned in the standup, I looked at the logs from the resources service and the behaviour is the same as when removing other fields, such as a phone number. The difference is that with identifiers you have the extra layer of `structuredIdentifier` and `localId`. The resources service does not delete the triple for the `localId` but only deletes and re-inserts the triple for `structuredIdentifier`.

app/models/identifier.js
aliokan: @mirdono I like the `/^\d*$/` solution! :+1:

aliokan (approved):
Approved :rocket:

PR #601, opened 2024-04-12
fix: Corrected typo in function call

mirdono:
This typo caused the match to fail for OCMW associations resulting in using the wrong link.


aliokan (approved):
Good catch! :+1: 

This kind of misspelling need to be catch automatically. I created a follow-up OP-3154

PR #604, opened 2024-04-25
refactor: move remaining classification code functionality into model

mirdono:
DO NOT MERGE: wait until #602 has been merged and tested

This PR moves the remaining classification code functionality that was still in some controllers to the `administrative-unit` model. The controllers for several forms (new organization, edit core data, and edit related organizations) all contained the same functionality to get a list of valid classification codes for participants and founders for several types of organizations.

Notes:
- The similar code in the `SiteTypeSelectComponent` was not modified as this component should be removed, along with other site edit code, once the CLB app takes over the editing of site information.
- The `municipalityClassificationCode` getter is rather stupid function to have. But I have left it in as an equivalent of the `classificationCodes` getter that was previously used. This getter would probably be removed if/when the relations between organizations are properly sorted out.


aliokan (approved):
Thank you! :pray:

PR #605, opened 2024-04-26
refactor: use "Representative body" instead of "Representative organ"

mirdono:
Previously both terms, body and organ, were used interchangeably. This commit
changes all uses of organ to body so that we have consistent terminology
throughout the frontend.

Body was chosen over organ as it is also used in the backend and we already use
the term "governing **body**" for "bestuurs**orgaan**".


aliokan (approved):
I see the same changes as https://github.com/lblod/frontend-organization-portal/pull/604/files, I guess it will not be a problem.

PR #607, opened 2024-05-03
fix: moved worship classification checks to administrative unit model

mirdono:
Bug reported by @cecemel that prevents the creation of new worship
services. Also reported by the testers (OP-3178) shortly after the initial
report.

The form for new organizations uses an administrative-unit record. This means
that if the functions are placed in the more specific models they cannot be
correctly used when deciding which fields to show.

More concretely, the fields for "Soort eredienst", "Strekking", and
"Grensoverschrijdend" were not shown when creating worship services. The
get-functions used to trigger displaying these fields were defined in the
worship service model and therefore not available in the form.

Note: While fixing this bug I noticed a similar problem with the "Centraal
bestuur" field that is never shown. This bug was introduced in earlier
refactoring and affects more forms. This will be fixed in a follow-up PR.

### DIFF
--- a/app/models/administrative-unit.js
+++ b/app/models/administrative-unit.js
@@ -261,4 +261,20 @@ export default class AdministrativeUnitModel extends OrganizationModel {
       this.isRepresentativeBody
     );
   }
+
+  // NOTE: the following worship-related functions have to be placed here
+  // instead of their corresponding models. The new organization form works
+  // based on an administrative unit record and does not show the correct
+  // fields if these functions are placed in the more specific models.
+  get isWorshipAdministrativeUnit() {
+    return this.isWorshipService || this.isCentralWorshipService;
+  }
+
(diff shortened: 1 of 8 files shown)


cecemel:
I am not sure i am the best to review; best to check with the team that works on this. Thanks for the fix!

aliokan (approved):
In the future, we'll need to reconsider how models are instantiated when create a new organization. Use "administrative unit" by default doesn't seem logical.

mirdono:
@aliokan You are 100% correct, while looking into/fixing this issue I found that our validation for new organizations is partially broken because of this. More specifically, the worship administrative-unit-specific validations (such as having a value for "Soort eredienst") are simply not checked/enforced. Only the validations defined in the administrative-unit and organization models are enforced. I will make a follow-up ticket for this.

PR #609, opened 2024-05-08
fix: perform worship-specific validations on organization creation

mirdono:
OP-3183

## Problem

Previously, the creation of organizations was done primarily using the plain `administrative-unit` model. Only after the validations succeeded on this model more specific models such as `worship-service` and `central-worship-service` were populated with the user input and store. Consequently, any validations defined in these child models (or in `worship-administrative-unit`) were never validated. This, for example, resulted in the possibility of creating a new `worship-service` without filling in the mandatory "Soort eredienst" field.

## Proposed solution

- In the router, instantiate the `administrativeUnit` model with an instance of either the `administrative-unit` or `worship-administrative-unit` model depending on the role of the user. This allows to automatically perform some additional validations for the worship module.
- For the worship module, populate all organizations models in the route with the values provided by the user. For example, setting only the name for the `administrativeUnit` model instance, this is now also set for the `worshipService` and `centralWorshipService` model instances. This allows to validate these model instances before saving them to the backend.

## TODO

- [ ] Check how well the applied solutions scale when introducing non-administrative unit organizations


aliokan:
@mirdono This what I have in mind: r

Rather than switching the model at the end of the flow, when user clicks on the save button. We can initially load the form with the most generic model type, `organization`. Then we can dynamically convert the model each time the user modifies the `organizationClassificationCode`. 

This method preserves the alignment of validation rules with the appropriate model, eliminating the need to handle exceptions in the router and controller. To achieve that we need to create a convert method that will take as arguments the current organization model and the `organizationClassificationCode`, and return the correct organization type model instance after copying all the current values of the previous one. 

```
organizationConverter(currentOrganizationModelInstance, organizationClassificationCode) : NewOrganizationModelInstance
```

I believe this approach offers a more simple and flexible solution to managing organization types within the creation form. What are your thoughts on this proposal?

mirdono:
I agree that using only a single model instance, which is converted when necessary, is a preferable way of solving this problem. I did initially attempt some things this direction but never got it fully working, either things broke outright or the form did not update correctly. I will give it another go with your suggestion in mind.

mirdono:
Superseded by #611

PR #610, opened 2024-05-14
Add message when kboOrganization is missing

aliokan:
## Before 

image

### Use case 1: KBO number is missing

image

###  Use case 2: KBO is not missing, but there is no information in Wegwijs

image

### DIFF
--- a/app/components/kbo-organization/data-card.js
+++ b/app/components/kbo-organization/data-card.js
@@ -1,0 +1,8 @@
+import Component from '@glimmer/component';
+
+export default class DataCardComponent extends Component {
+  noKboMessage =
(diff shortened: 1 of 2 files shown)


mirdono (approved):
Looks good to me

app/components/kbo-organization/data-card.js
aliokan, line 1: @mirdono I put messages here because they are tied to the component. I'm not convinced it's worth polluting the `help-text` with them.

mirdono: I was doubting about that for a moment indeed. But then came to same conclusion as you :-)

PR #611, opened 2024-05-16
Generalise organisation creation to operate on the most specific model type

mirdono:
## Summary

Previously the functionality for creating new organisations mainly operated on the `AdministrativeUnitModel`. This PR makes that functionality more dynamic such that it always uses the most specific organisation model type depending on the classification selected by the user.

## Related tickets:

- OP-3183: this refactoring brings also a better fix for that bug than was proposed in #609
- OP-3026: these changes should make it easier to extend the organisation creation functionality to non-administrative unit organisations

## Previous situation

The organisation model instances available in the new organisation route were create a priori by the router. Any input provided by the user was assigned to the generic `AdministrativeUnitModel` instance. Only after this model was successfully validated were the user inputs copied to one of the more specific models, such as the `WorshipModel` or `CentralWorshipModel` instances.

One consequence of this flow was that validations specific to the more specific types of models were never checked. It was, for example, possible to create a new worship service without providing a value for its mandatory recognised worship type property (the "Soort eredienst" input field). This was documented in OP-3183.

Another consequence of this flow is that it is difficult to adapt to allow the creation of non-administrative unit organisations in the near future.

## Proposed solution

The model instances are now created on the fly by the controller, taking into account the classification selected by the user. Initially the router instantiates the model variable with an instance of the most general `OrganizationModel`.
Each time the user changes to classification a new model instance is created of the most specific model type. Any necessary properties are copied to the new model from the previous model. Any relationships between the organisation being created and already existing organisations are **not** copied. These relations tend to be specific to certain classifications and should be explicitly set (again) by the user.

Furthermore, some simple properties are only copied between organisations with compatible classifications. More specifically, only organisations that are an IGS[1] can have an expected end date and/or purpose. Previously, these properties were, when set, copied to the saved model, irrelevant of the organisation's final classification. This resulted in the possibility to create non-IGS organisations with an expected end date and/or purpose by

1. Selecting an IGS classification
2. Entering a value in the fields for expected end date and/or purpose
3. Changing the classification to a different value (non-IGS)
4. Saving the model

## TODO

- [X] The call to `deleteRecord()` sometimes causes an `InternalError` concerning too much recursion (exact error message differs per web browser). Further investigation revealed this seems to be an issue with the Ember Inspector plugin. So far I have been unable to reproduce the error when the plugin is not opened in the developer console.
- [x] Worship: Clear the relationships also when user changes the recognised worship type. Otherwise it is possible to store a worship service that is sub organisation of a central worship service for worships types that don't have this.

[1] A group of 4 classifications: "Projectvereniging", "Dienstverlenende vereniging", "Opdrachthoudende vereniging", and "Opdrachthoudende vereniging met private deelname"

### DIFF
--- a/app/controllers/organizations/new.js
+++ b/app/controllers/organizations/new.js
@@ -106,12 +106,12 @@ export default class OrganizationsNewController extends Controller {
       let newOrganizationModelInstance = this.#createNewModelInstance(
         organizationClassificationCode.id
       );
-      // Note: explicitly set here to ensure form is updated
-      newOrganizationModelInstance.classification =
-        organizationClassificationCode;
 
       // Copy attributes and relationships to new model instance
       this.#copyPropertiesToModel(newOrganizationModelInstance);
+      // Note: explicitly set here to ensure form is updated
+      newOrganizationModelInstance.classification =
+        organizationClassificationCode;
 
       // Delete the old model instance
       // Note: this sometimes causes an InternalError concerning too much
--- a/app/controllers/organizations/new.js
+++ b/app/controllers/organizations/new.js
@@ -1,19 +1,23 @@
 import Controller from '@ember/controller';
 import { inject as service } from '@ember/service';
 import { dropTask } from 'ember-concurrency';
+import { tracked } from '@glimmer/tracking';
 import { combineFullAddress } from 'frontend-organization-portal/models/address';
 import { action } from '@ember/object';
 import { setEmptyStringsToNull } from 'frontend-organization-portal/utils/empty-string-to-null';
 import fetch from 'fetch';
 import { transformPhoneNumbers } from 'frontend-organization-portal/utils/transform-phone-numbers';
+import { CLASSIFICATION } from '../../models/administrative-unit-classification-code';
 
 export default class OrganizationsNewController extends Controller {
   @service router;
   @service store;
 
+  @tracked currentOrganizationModel;
+
   get hasValidationErrors() {
     return (
-      this.model.administrativeUnit.error ||
+      this.currentOrganizationModel.error ||
       this.model.address.error ||
       this.model.contact.error ||
       this.model.secondaryContact.error ||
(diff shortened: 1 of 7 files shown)


app/controllers/organizations/new.js
```diff
@@ -47,43 +51,161 @@ export default class OrganizationsNewController extends Controller {
...
+    if (
+      organizationClassificationCode.id !==
+      this.currentOrganizationModel.classification.id
+    ) {
+      // Create new model instance based on the provided classification
+      let newOrganizationModelInstance = this.#createNewModelInstance(
+        organizationClassificationCode.id
+      );
+      // Note: explicitly set here to ensure form is updated
+      newOrganizationModelInstance.classification =
```
aliokan, line 110: **Nitpick:** I would move that line after the `copyPropertiesToModel` to avoid the classification to be overrided in `copyPropertiesToModel`

app/controllers/organizations/new.js
```diff
@@ -47,43 +51,161 @@ export default class OrganizationsNewController extends Controller {
...
+    }
+
+    // Only IGSs have an, optional, expected end data and/or goal. Only copy
+    // these values when new organization is also an IGS.
+    if (newOrganizationModel.isIgs) {
+      newOrganizationModel.expectedEndDate =
+        this.currentOrganizationModel.expectedEndDate;
+      if (this.currentOrganizationModel.purpose?.length) {
+        newOrganizationModel.purpose = this.currentOrganizationModel.purpose;
+      }
```
aliokan, line 209: **Question:** Is it normal to have less properties copied in the new function than in the original one (eg. `wasFoundedByOrganizations`, `isAssociatedWith`...)? https://github.com/lblod/frontend-organization-portal/pull/611/files#diff-85ba2ee0843881ee36cae41399c083e5cd4141e9ad065e1cc2ca28c30edce447L205-L246

aliokan (commented):
Looks good to me. I only left 2 small comments

app/controllers/organizations/new.js
aliokan: I found the answer in the PR description, you can forget that comment ;)

aliokan (approved):

app/controllers/organizations/new.js
mirdono: That is indeed the more robust order of doing these things.

PR #614, opened 2024-05-22
feature: add functionality for private OCMW associations

mirdono:
OP-3026

The information shown to the user is the same as for the public OCMW
associations (see #553 and OP-2540). The primary difference is that for private
OCMW associations we do **not** include functionaries ("Leidinggevenden") and
therefore not synchronise these with Loket.

This should be tested against the backend PR importing the corresponding
business data: https://github.com/lblod/app-organization-portal/pull/426

## Main changes:
- Uncommented/added the necessary classification codes:
  + "Ziekenhuisvereniging"
  + "Vereniging of vennootschap voor sociale dienstverlening"
  + "Woonzorgvereniging of woonzorgvennootschap"
- Added a new model for registered organisations
- Extending the new organisation controller to instantiate a registered
  organisation model when an appropriate classification is selected


aliokan (approved):
looks good! :rocket:

PR #617, opened 2024-05-24
Add missing `representative-body-classification-codes` model

aliokan:
Related to OP-3229

## Context 

The representative body classification codes was recently fixed in the triplestore by https://github.com/lblod/app-organization-portal/pull/412

As a side effect, a code typed `representative-body-classification-codes` is part of `/organization-classification-codes` response. The app is complaining because the model is missing. 

## Fix

Adding the missing model.


mirdono (approved):
Looks good, thanks for fixing this!

PR #623, opened 2024-05-31
fix: correctly include region when editing a PEVA

mirdono:
Previously the region was not loaded in the model for PEVAs and the field was
not shown in the core data edit form. These changes correct this to align the
behaviour with the new organisation form as well as the core data editing for
other kinds of organizations.


aliokan (approved):
Good catch! :stuck_out_tongue_winking_eye:

PR #621, opened 2024-05-28
Add edit contact behind feature flag in new addministrative-unit form

aliokan:
Related to OP-3248

Before
image

After
image

PR #624, opened 2024-06-03
Fix/addministrative unit without address and contact

aliokan:
Related to OP-3251, Follow-up of https://github.com/lblod/frontend-organization-portal/pull/621

- Remove contact and address validations in new/edit administrative-unit form when edit flag is off
- Handle administrative-unit without address and contact

### DIFF
--- a/app/controllers/administrative-units/administrative-unit/core-data/edit.js
+++ b/app/controllers/administrative-units/administrative-unit/core-data/edit.js
@@ -135,60 +135,67 @@ export default class AdministrativeUnitsAdministrativeUnitCoreDataEditController
 
     yield Promise.all([
       administrativeUnit.validate({ relaxMandatoryFoundingOrganization: true }),
-      address.validate(),
-      contact.validate(),
-      secondaryContact.validate(),
       identifierKBO.validate(),
       identifierSharepoint.validate(),
     ]);
 
+    if (this.features.isEnabled('edit-contact-data')) {
+      yield Promise.all([
+        address.validate(),
+        contact.validate(),
+        secondaryContact.validate(),
+      ]);
+    }
+
     if (!this.hasValidationErrors) {
-      let primarySite = yield administrativeUnit.primarySite;
+      if (this.features.isEnabled('edit-contact-data')) {
+        let primarySite = yield administrativeUnit.primarySite;
 
-      // TODO : "if" not needed when the data of all administrative units will be correct
-      // they should all have a primary site on creation
-      if (!primarySite) {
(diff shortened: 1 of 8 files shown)


app/controllers/administrative-units/new.js
```diff
@@ -144,13 +145,18 @@ export default class AdministrativeUnitsNewController extends Controller {
...
+    if (this.features.isEnabled('edit-contact-data')) {
+      yield Promise.all([
+        address.validate(),
+        contact.validate(),
+        secondaryContact.validate(),
+      ]);
+    }
+
     if (!this.hasValidationErrors) {
       const siteTypes = yield this.store.findAll('site-type');
```
mirdono, line 161: **Nitpick:** I assume this call can move into the conditional `edit-contact-data` block. (Not sure why it is called all the way up here.)

app/templates/administrative-units/administrative-unit/core-data/edit.hbs
```diff
@@ -346,7 +346,9 @@
               Primaire contactgegevens 
             </:title>
             <:infotext>
-               De primaire contactgegevens kunnen niet langer in het OP aangepast worden. De organisaties kunnen deze gegevens zelf aanpassen via de Contactapp.
```
mirdono, line 349: **Comment:** Imo the resulting page looks a bit weird for an organisation without an address. If the organisation has no address yet the data card shows an empty box underneath the message:
2024-06-03-184013_995x114_scrot

Could we make in a bit more obvious that an address has just not been added yet?

app/controllers/administrative-units/administrative-unit/core-data/edit.js
```diff
@@ -135,60 +135,67 @@ export default class AdministrativeUnitsAdministrativeUnitCoreDataEditController
...
+      ]);
+    }
+
     if (!this.hasValidationErrors) {
-      let primarySite = yield administrativeUnit.primarySite;
-
-      // TODO : "if" not needed when the data of all administrative units will be correct
-      // they should all have a primary site on creation
-      if (!primarySite) {
-        primarySite = primarySite = this.store.createRecord('site');
```
mirdono, line 151: **Nitpick:** I guess one `primarySite = ` is sufficient here.

mirdono (approved):
Some minor comments and nitpicks, nothing blocking for me.

app/templates/administrative-units/administrative-unit/core-data/edit.hbs
aliokan: I agree, I will deploy and see what Sofi think about

PR #626, opened 2024-06-06
refactor: resolved fixme central worship service field

mirdono:
Related to OP-3177 and OP-3183

In #608 a dirty hack was applied to resolve a bug were the central worship
service field was not shown when appropriate. Now that the new organisation
handles the different model types in a proper manner this (see #611) this quick
fix can be replaced by the proper function call.


aliokan (approved):
I like that! :pray:

PR #627, opened 2024-06-06
feat: ignore edit-contact-data flag for private OCMW associations

mirdono:
OP-3250

Private OCMW associations are registered organisations for which the CLB app
will not become the master of their address and contact data.

## Proposed solution
- In controllers and templates, add a check whether the organisation is a
  private OCMW association. This check essentially allows to ignore the feature
  flag.
- In routes, move the check on the feature flag to an `afterModel` hook and add
  a similar check as for controller template overwriting the value of the
  feature flag.


aliokan (approved):
Looks good! :+1:

PR #628, opened 2024-06-10
feat: add associations and corporations

mirdono:
OP-3023

## Summary
Add new types of organizations to OP: associations (*nl. verenigingen*) and
corporations (*nl. vennootschappen*), both types are registered
organizations. This includes functionality to create these new types of
organizations and define relationships with other organizations. Furthermore, an
new search/filter field was added to the organizations index page to filter
based on type of organization. The changes should **not** impact the worship module.

### Creating new organizations
A new field was added to the form for creating new organizations. The intended
flow is that a user now first selects an organization type (administrative unit,
association, or corporation). Second, the user has to select the classification
of the new organization, here the options are limited to those matching the
organization type. Otherwise the form should function as before.

For the new types of organizations the user has to provide address and contact
data, OP remains the master of this data for non-administrative unit
organizations.

For the new types of organizations, the user cannot specify relationships with other organizations via this form. In other words, the related organizations block is shown for associations or corporations. The possible relationships should be entered on the side of the organizations in which is participated or was founded (see jira ticket for more details)

Open discussion point:
- When selecting association or corporation as organization type, a
  classification with the same label is available. This is possibly confusing
  for the user. An alternative would be to make the classification field
  optional for the new organization types and silently fill the classification
  when left blank.

### Organizations index page
A new filter was added to the organizations overview page, this filter allows to
limit the displayed organizations to specific types. When no organization type
is selected everything is shown, otherwise the table is limited to the select
organization types (this is the same behaviour as the already existing
classification filter.)

This new filter interacts with the already existing classification filter:
- when one or more organization types are selected, the possible classifications
  are limited to those matching these organization types; and
- when one or more classifications are selected, the possible organization
  types are limited to those matching these classifications.

## Related PRs
- backend: TODO
- construct-organization-relationships-service: https://github.com/lblod/construct-organization-relationships-service/pull/14


app/controllers/organizations/new.js
```diff
@@ -79,6 +100,31 @@ export default class OrganizationsNewController extends Controller {
     this.currentOrganizationModel.isSubOrganizationOf = null;
   }
 
+  @action
+  updateOrganizationType(organizationType) {
+    // Note: organization type is *not* used for worship module
+    if (
+      this.currentSession.hasUnitRole &&
```
aliokan, line 107: **Question:**Is it needed?

app/controllers/organizations/new.js
```diff
@@ -89,92 +135,67 @@ export default class OrganizationsNewController extends Controller {
...
-        CLASSIFICATION.POLICE_ZONE.id,
-        CLASSIFICATION.ASSISTANCE_ZONE.id,
-        CLASSIFICATION.WELZIJNSVERENIGING.id,
-        CLASSIFICATION.AUTONOME_VERZORGINGSINSTELLING.id,
-        CLASSIFICATION.PEVA_MUNICIPALITY.id,
-        CLASSIFICATION.PEVA_PROVINCE.id,
-      ].includes(classificationCodeId)
-    ) {
-      return this.store.createRecord('administrative-unit');
+    if (this.currentSession.hasUnitRole && this.organizationType) {
```
aliokan, line 180: **Comment:** Is `this.currentSession.hasUnitRole` needed?

app/controllers/organizations/new.js
```diff
@@ -89,92 +135,67 @@ export default class OrganizationsNewController extends Controller {
...
-   * @param {string} classificationCodeId - the unique identifier of
-   * the selected classification code
+   * @param {string} [classificationCodeId] - the unique identifier of the
+   *     selected classification code.
    * @returns {OrganizationModel} New model instance that matches to provided
-   * classification code
+   *     classification code.
    */
   #createNewModelInstance(classificationCodeId) {
-    // FIXME: This logic is somewhat duplicate with the classification getters
```
aliokan, line 139: **Suggestion:** In order to simplify, instead of all these if/else, can we move the `CLASSIFICATION` to model link in the `app/models/administrative-unit-classification-code.js` configuration file.

app/controllers/organizations/new.js
```diff
@@ -240,10 +269,13 @@ export default class OrganizationsNewController extends Controller {
...
       identifierSharepoint.validate(),
     ]);
+    this.attemptedSave = true;
 
     if (
       this.features.isEnabled('edit-contact-data') ||
-      this.currentOrganizationModel.isPrivateOcmwAssociation
+      this.currentOrganizationModel.isPrivateOcmwAssociation ||
+      this.currentOrganizationModel.isAssociationOther ||
+      this.currentOrganizationModel.isCorporationOther
```
aliokan, line 278: **Suggestion:** Instead of redefining the condition every time, can we create a utils function `isContactEditableOrganization`.

app/templates/organizations/new.hbs
```diff
@@ -282,6 +312,8 @@
       (or
         (is-feature-enabled "edit-contact-data")
         this.currentOrganizationModel.isPrivateOcmwAssociation
+        this.currentOrganizationModel.isAssociationOther
+        this.currentOrganizationModel.isCorporationOther
```
aliokan, line 316: Same here

aliokan (changes requested):
Let discuss with Sofi about the new `organization type` field during the standup

app/controllers/organizations/new.js
mirdono: Let's keep this for a follow up as it is a broader refactoring than just for this form.

aliokan (approved):
looks good! :+1:

app/controllers/organizations/organization/core-data/edit.js
```diff
@@ -5,6 +5,7 @@ import { combineFullAddress } from 'frontend-organization-portal/models/address'
 import { action } from '@ember/object';
 import { setEmptyStringsToNull } from 'frontend-organization-portal/utils/empty-string-to-null';
 import { transformPhoneNumbers } from 'frontend-organization-portal/utils/transform-phone-numbers';
+import isContactEditableOrganization from '../../../../utils/editable-contact-data';
```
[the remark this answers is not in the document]
mirdono: Yeah, I tend to favour using absolute paths, but I have not yet been able to convince my language server, which inserted these import statements, of that.

PR #631, opened 2024-06-21
refactor: unify related organizations components

mirdono:
DO NOT MERGE: this PR is meant as preparation for the tickets part of OP-3195,
only merge to development once v1.26.0 is on QA

Related to epic OP-3195

This refactoring merges the different, often highly similar, components for
viewing and editing related organizations into a smaller number of
components. The goal is to:
- reduce the amount of code duplication in this part of the frontend
- simplify the further code changes needed for the reworking of related
  organizations as documented in the mentioned epic.
- keep the functionality equivalent to how it was. Note that small differences
  are present, for example the order of elements in the data card for some
  organizations classifications has changed.


aliokan (approved):
looks good! :pray:

PR #634, opened 2024-08-13
[OP-3325] Add whitespace between each alternative name item

Windvis:
An organization can have multiple alternative names. We display this list in a template directly, which causes the VM to call `.toString` on it which defaults to joining the array elements without any whitespace.

This adds a new `join-array` helper that can use different separators to join the array.

### DIFF
--- a/app/templates/organizations/organization/core-data/index.hbs
+++ b/app/templates/organizations/organization/core-data/index.hbs
@@ -82,7 +82,7 @@
                 <Item>
                   <:label>Alternatieve naam</:label>
                   <:content>
-                    {{@model.organization.alternativeName}}
+                    {{join-array @model.organization.alternativeName}}
                   </:content>
                 </Item>
               {{/if}}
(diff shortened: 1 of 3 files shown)


app/templates/organizations/organization/core-data/index.hbs
```diff
@@ -82,7 +82,7 @@
                 <Item>
                   <:label>Alternatieve naam</:label>
                   <:content>
-                    {{@model.organization.alternativeName}}
+                    {{join-array @model.organization.alternativeName}}
```
Windvis, line 85: Alternatively we could add a getter to the `Organization` class which would do the formatting, and use that here instead, but a generic helper seemed better.

mirdono: Agreed, the helper seems the better solution. In retrospect, I should have already added this back when introducing the alternative name functionality /Entered on/ [2024-08-13 Tue 10:06]

mirdono (approved):
Looks good, thanks for fixing this

PR #635, opened 2024-08-14
[OP-3334] Dependency maintenance

Windvis:
This updates (or removes, if they aren't needed anymore) most of the dependencies that aren't included in the app blueprint and didn't require substantial code changes.


app/routes/application.js
```diff
@@ -6,9 +6,12 @@ export default class ApplicationRoute extends Route {
   @service session;
   @service role;
   @service router;
+  @service intl;
 
   async beforeModel() {
     await this.session.setup();
+    this.intl.setLocale(['nl-BE']);
```
Windvis, line 13: This works around an issue with ember-intl v6, but it's good to do anyways since we want the editor labels to be in Dutch.

PR #636, opened 2024-08-14
[OP-3343] Appuniversum v3 update

Windvis:
This resolves all v2 deprecations following the [deprecation guide](https://github.com/appuniversum/ember-appuniversum/pull/459) and updates to Appuniversum v3.

**Notes:**
Appuniversum v3 introduced a new `AuLoader` deprecation, but this PR doesn't resolve that yet since it doesn't break anything and I didn't want to scope creep the PR.

PR #637, opened 2024-08-20
[OP-3042] Resolve ember v4 deprecations

Windvis:
This resolves any EmberData related deprecations I could find.

I clicked through the app as best as I could, but I'm unfamiliar with the app so I might have missed some flows.

To make testing easier, you could use the following deprecation-workflow config to hide the noisy non-EmberData deprecations:

<details>
<summary>Config</summary>

```js
self.deprecationWorkflow.config = {
  workflow: [
    // TODO: resolve before updating to Appuniversum v4
    {
      handler: 'silence',
      matchId: '@appuniversum/ember-appuniversum.au-loader-visible-by-default',
    },
    // TODO: resolve before updating to Appuniversum v4
    {
      handler: 'silence',
      matchId: '@appuniversum/ember-appuniversum.au-button-loading-message',
    },
    { handler: 'silence', matchId: 'ember-resources.util.trackedTask' }, // TODO: Resolve before updating ember-resources
    { handler: 'silence', matchId: 'ember-resources.class-based' }, // TODO: Resolve before updating ember-resources
  ],
};
```
</details>

~~(Builds on #636 so draft until that's merged)~~

### DIFF
--- a/app/models/membership-role.js
+++ b/app/models/membership-role.js
@@ -40,11 +40,13 @@ export default class MembershipRoleModel extends Model {
   @attr('string') label;
 
   @belongsTo('concept', {
+    async: true,
     inverse: null,
   })
   topConceptOf;
 
   @belongsTo('membership-role', {
+    async: true,
     inverse: null,
   })
   hasBroaderRole;
--- a/app/templates/organizations/organization/change-events/new.hbs
+++ b/app/templates/organizations/organization/change-events/new.hbs
@@ -304,7 +304,7 @@
                     {{else}}
                       <div class={{if hasError "ember-power-select--error"}}>
                         <PowerSelect
-                          @options={{@model.changeEvent.originalOrganizations}}
+                          @options={{has-many-value @model.changeEvent "originalOrganizations"}}
                           @selected={{this.selectedResultingOrganization}}
                           @onChange={{this.updateResultingOrganizations}}
                           as |organization|
--- a/tests/unit/models/membership-test.js
+++ b/tests/unit/models/membership-test.js
@@ -68,8 +68,8 @@ module('Unit | Model | membership', function (hooks) {
             role: relationRole,
           });
 
-          organization.memberships.pushObject(participant);
-          organization.memberships.pushObject(model);
+          (await organization.memberships).push(participant);
+          (await organization.memberships).push(model);
 
           const isValid = await model.validate({
             creatingNewOrganization: true,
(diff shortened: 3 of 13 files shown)


tests/unit/models/membership-test.js
```diff
@@ -109,7 +109,7 @@ module('Unit | Model | membership', function (hooks) {
             role: role,
           });
 
-          organization.memberships.pushObject(model);
+          organization.memberships.push(model);
```
Windvis, line 112: Some of these `.push` calls seem to fail after updating to v5, but the tests passed on v4. I need to double check that.

Windvis:
The [issue mentioned here](https://github.com/lblod/frontend-organization-portal/pull/638#issuecomment-2298947325) is also relevant in this PR, so we should fix it here.

app/templates/organizations/organization/change-events/new.hbs
```diff
@@ -304,7 +304,7 @@
                     {{else}}
                       <div class={{if hasError "ember-power-select--error"}}>
                         <PowerSelect
-                          @options={{@model.changeEvent.originalOrganizations}}
+                          @options={{has-many-value @model.changeEvent "originalOrganizations"}}
```
Windvis, line 307: I went with a new helper since we might need it in other places that I haven't encountered yet.

I decided not to go for EmberData's `Await` component or ember-promise-helpers since the data is already loaded, we simply want sync access to the already loaded data.

mirdono (commented):
I have left some comments/questions based on a first look.

app/models/membership-role.js
```diff
@@ -40,11 +40,13 @@ export default class MembershipRoleModel extends Model {
   @attr('string') label;
 
   @belongsTo('concept', {
+    async: true,
     inverse: null,
   })
   topConceptOf;
 
   @belongsTo('membership-role', {
+    async: true,
```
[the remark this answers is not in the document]
mirdono: I took another look at it and it seems I ran into a bug that was already present, so it has nothing to do with this change. Apologies for pointing fingers too quickly :blush: 

For completeness: the bug occurs when adding and removing a new related organization without filling in the data. In that case when pressing save a `TypeError` occurs because, presumably, the removal of the empty membership does not work correctly. Bug logged as OP-3371 which we'll resolve as part of the whole membership refactoring.

mirdono (approved):
I made another pass and imo we are good to go with this one. The points I raised previously were resolved or turned out to be unrelated.

The remaining deprecations I encountered while going through are outside the scope of this PR.

PR #638, opened 2024-08-20
[OP-3042] Ember v5 update

Windvis:
This updates the app to the Ember v5.8 LTS blueprint and resolves the linting errors and deprecations I encountered. 

(Builds on #637 so draft until that is merged)


Windvis:
An issue I encountered by clicking around some more:

1. add a new change event
2. choose "Samenvoeging"
3. error will be thrown in the console (most likely a deprecation in the v4 branch)

Windvis:
Note: I noticed the following deprecation: https://deprecations.emberjs.com/id/ember-data-deprecate-non-uniq-collected-payloads/ while clicking through the app, but it's not very clear if or how it should be resolved. The description of the deprecation makes it seem like this is an API issue, but pushing the same record in a hasMany relationship multiple times also triggers it.

We would have to do presence checks before pushing, which just seems silly to me. I'm hoping they make this deprecation less strict and only deprecate the behavior when processing API responses.

I asked in the Discord, but thus far I haven't received a response yet.

PR #641, opened 2024-08-23
fix: correctly calculate total sum of finance percentages for local involvements

mirdono:
The validation that the total sum of financing percentages for local involvements is exactly 100 contained an error where any newly added local involvements were not considered in the calculation.

## Steps to reproduce the original bug
1. Log in as worship editor
2. Select a worship service in the organisation overview
3. Navigate to the “Betrokken lokale besturen” page
4. Click the “Bewerk” button in the top right corner
5. Add a new local involvement by clicking the “+ Voeg nieuw betrokken lokaal bestuur toe” button on the bottom
6. Fill in the necessary fields: select “Mede-financierend” as type of involvement, enter a percentage above 0
7. Make sure the total sum of the percentages equals 100
8. Click the save button in the top right corner
9. On most percentage fields an error should be shown concerning an incorrect total percentage sum (Full error message: 'Het totaal van alle percentages moet gelijk zijn aan 100')

## Problem description
In order to calculate the sum of the finance percentages in the local involvement model validation, the other relevant local involvements must be retrieved. Initially this was done with a query on the store for local involvements, filtered on the identifier of the worship administrative unit related to the local involvement being validated.

This way of retrieving local involvements resulted in that any newly added, unsaved local involvement was not retrieved and thus not included in the sum calculation.

## Proposed solution
Instead of querying the store, we can use the `involvements` relationship of the worship administrative unit related to the local involvement being validated. This relationship includes all previous as well as newly added local involvements. Resulting in a correct calculation of the total financing percentage.

## Related tickets
OP-3373

### DIFF
--- a/app/models/local-involvement.js
+++ b/app/models/local-involvement.js
@@ -80,6 +80,7 @@ export default class LocalInvolvementModel extends AbstractValidationModel {
           .external(async (value, helpers) => {
             const otherLocalInvolvements =
               await this.getOtherLocalInvolvements();
+
             const sumOtherPercentages = otherLocalInvolvements.reduce(
               (percentageAcc, involvement) =>
                 percentageAcc + Number(involvement.percentage),
@@ -144,14 +145,11 @@ export default class LocalInvolvementModel extends AbstractValidationModel {
    */
   async getOtherLocalInvolvements() {
     const worshipAdministrativeUnit = await this.worshipAdministrativeUnit;
-    let relatedLocalInvolvements = await this.store.query('local-involvement', {
-      filter: {
-        'worship-administrative-unit': {
-          id: worshipAdministrativeUnit.get('id'),
-        },
-      },
-    });
+    const relatedLocalInvolvements =
+      await worshipAdministrativeUnit.involvements;
 
-    return relatedLocalInvolvements.filter((elem) => elem.id !== this.id);
+    // Note: do not filter on id as this also deals with unsaved elements that
+    // do not have an id yet.
+    return relatedLocalInvolvements.filter((elem) => elem !== this);
   }
 }
--- a/tests/unit/models/local-involvement-test.js
+++ b/tests/unit/models/local-involvement-test.js
@@ -102,6 +102,150 @@ module('Unit | Model | local involvement', function (hooks) {
       });
     });
 
+    test('it does not return an error when the percentage is 100', async function (assert) {
+      const administrativeUnit = this.store().createRecord(
+        'administrative-unit'
+      );
+      const involvementType = this.store().createRecord('involvement-type', {
+        id: INVOLVEMENT_TYPE.SUPERVISORY,
+      });
+      const model = this.store().createRecord('local-involvement', {
+        administrativeUnit,
+        involvementType,
+        percentage: 100,
+      });
+
+      const getOtherLocalInvolvementsStub = sinon.stub(
+        model,
+        'getOtherLocalInvolvements'
+      );
+      getOtherLocalInvolvementsStub.resolves([]);
+
+      const isValid = await model.validate();
+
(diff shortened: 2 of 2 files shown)


Riadabd (approved):
I verified first the bug existed using your method in the description, and then confirmed it's fixed with this PR.

Percentages for newly added `betrokken lokaal besturen` are correctly picked up, and the error shows up as expected if percentage is > or < 100. :+1:

app/models/local-involvement.js
```diff
@@ -144,13 +145,8 @@ export default class LocalInvolvementModel extends AbstractValidationModel {
...
     const worshipAdministrativeUnit = await this.worshipAdministrativeUnit;
-    let relatedLocalInvolvements = await this.store.query('local-involvement', {
-      filter: {
-        'worship-administrative-unit': {
-          id: worshipAdministrativeUnit.get('id'),
-        },
-      },
-    });
+    const relatedLocalInvolvements =
+      await worshipAdministrativeUnit.involvements;
```
Windvis, line 149: I'm not sure if this is safe, since relationships are paginated by mu-cl-resources AFAIK. Doesn't this break if there are more than 20 records in that relationship?

What was the issue with the previous code?

Riadabd: After seeing your comment, I tried testing again and found a small issue.

If all existing percentages add up to 99% for example, adding a new instance having a 1% value and saving will work correctly. However, assume all existing percentages add up to 98%, you add two instances with each having a value of 1%, saving will lead to an error even though the actual count is equal to 100%.

If I try and print the calculated percentages, I can see `100%` but also `99%`, which I think is because of each new unsaved instance not being aware of the other, and thus not taking their value into account.

mirdono: @Windvis I had not considered that, will have to test this to make sure whether it is a issue.

@Riadabd, good catch. I think your assessment that the new instances are not aware of each other is correct. Will have to take a deeper look to see how to solve this.

I converted this PR to draft to avoid it is accidentally merged before the issues pointed out are resolved.

Windvis: > @Windvis I had not considered that, will have to test this to make sure whether it is a issue.

@mirdono I think it might work if the relationship data was already loaded before (with includes or a similar query as before) which is probably the case since the table is displayed already. I think EmberData returns everything that it has in the store. But if the data is fetched from the API, I think it's paginated.

Looking at it some more, I'm not the biggest fan of the validation setup to be honest :sweat_smile: . If the table contains 10 local involvements, doesn't that mean this logic runs 10 times as well? It feels wrong to hard wire form data and validations to the model classes like this.

Wouldn't it be better to do the "total percentage must be 100%" check somewhere else?

mirdono: @Windvis The local involvements in this route are indeed included when [retrieving the record](https://github.com/lblod/frontend-organization-portal/blob/80b24c585870a4830d2656c45f9c95707568128b/app/routes/organizations/organization/local-involvements/edit.js#L19) in this route. To be sure I manually checked by making 30 entries and the edit page does not paginate, the display page does paginate per 20 entries. (Sidenote, in production the highest number of local involvements for a single worship service is 9.)

You are also correct that the validation is needlessly executed once for each local involvement individually (which all have the same outcome if everything works correctly). Ideally, this validation would be performed only once at the level of the organisation or so. I do not recall exactly why I implemented it the way it is. Possibly to avoid validating the whole organisation, which can result in errors irrelevant to the form at hand, and/or because this way was easier to map error message to their input fields.

Note: this bug is present on PROD at the moment, so I would prefer first fixing the validation in the state that it is. Improving the actual implementation of the validation can be done "later", with the risk that it is postponed indefinitely...

Windvis (approved):

PR #644, opened 2024-09-05
fix: loosen up membership validations for vereniging en vennootschap

wolfderechter:
## ID
OP-3388

## Description

Fixes the bug when creating new verenigingen or vennootschappen of failing validations. The memberships validations was too strict and was mandatory for every type, but since verenigingen and vennootschappen don't need it, we allow the memberships to be empty then.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Other

## How to test

1. Create 2 new organisations:
- vennootschap
- vereniging
2. Fill in the required fields
3. Save and verify that the organisations get created successfully without the validations failing.

### DIFF
--- a/app/models/registered-organization.js
+++ b/app/models/registered-organization.js
@@ -27,7 +27,7 @@ export default class RegisteredOrganizationModel extends OrganizationModel {
       memberships: Joi.when(Joi.ref('$creatingNewOrganization'), {
         is: Joi.exist().valid(true),
         then: Joi.when('classification.id', {
-          is: Joi.string().valid(
+          is: Joi.exist().valid(
             ...CorporationOtherCodeList,
             ...AssociationOtherCodeList,
           ),
--- a/app/models/registered-organization.js
+++ b/app/models/registered-organization.js
@@ -29,7 +29,7 @@ export default class RegisteredOrganizationModel extends OrganizationModel {
         then: Joi.when('classification.id', {
           is: Joi.string().valid(
             ...CorporationOtherCodeList,
-            ...AssociationOtherCodeList
+            ...AssociationOtherCodeList,
           ),
           then: validateHasManyOptional(),
           otherwise: validateHasManyNotEmptyRequired(REQUIRED_MESSAGE),
--- a/app/models/registered-organization.js
+++ b/app/models/registered-organization.js
@@ -16,13 +16,24 @@ export default class RegisteredOrganizationModel extends OrganizationModel {
   get validationSchema() {
     const REQUIRED_MESSAGE = 'Selecteer een optie';
     return super.validationSchema.append({
-      // NOTE:The requested functionality was to *not* validate memberships of
-      // already existing organizations. When creating a new organization a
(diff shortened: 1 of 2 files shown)


app/models/registered-organization.js
```diff
@@ -16,13 +16,24 @@ export default class RegisteredOrganizationModel extends OrganizationModel {
...
+      // 2. For new organizations:
+      //    a. 'Vennootschappen' and 'Verenigingen' (corporations and associations):
+      //       memberships are optional.
+      //    b. All other types: memberships are mandatory.
+      // The creatingNewOrganization flag (set to true) triggers validation for new organizations.'
       memberships: Joi.when(Joi.ref('$creatingNewOrganization'), {
         is: Joi.exist().valid(true),
-        then: validateHasManyNotEmptyRequired(REQUIRED_MESSAGE),
+        then: Joi.when('classification.id', {
+          is: Joi.string().valid(
```
mirdono, line 30: **Issue:** I would replace `string()` with `exist()`, as we do in other places.

When the `classification` for a model is `undefined` the `string()` function behaves (imo) unexpected. In such a case the `is` clause succeeds as a whole, thereby performing the validation specified in its `then` branch. This is what causes the test to fail, as it takes the inner `then: validateHasManyOptional()` and longer results in a error for lacking memberships.



**Suggestion:** Add a test for the new situation.

Could you also add an additional test in the `validate` module of [`registered-organization-test`](https://github.com/lblod/frontend-organization-portal/blob/81d016f6f5b64f08dc5e975aced98a3d69ed295f/tests/unit/models/registered-organization-test.js#L13) to capture this changed validation situation for these two classes of organizations? These validations can be tricky when they need to be changed, so I prefer to have a sufficient set of tests that can indicate when a change has other effects. Something like the following should do the trick:
```js
[
  CLASSIFICATION.ASSOCIATION_OTHER,
  CLASSIFICATION.CORPORATION_OTHER,
].forEach((cl) => {
   test(`it should not return an extra error creating an empty ${cl.label} model`, async function (assert) {
     const classification = this.store().createRecord(
       'registered-organization-classification-code',
       cl,
     );
     const model = this.store().createRecord('registered-organization', {
       classification,
     });

     const isValid = await model.validate({ creatingNewOrganization: true });

     assert.false(isValid);
     assert.strictEqual(Object.keys(model.error).length, 2);
     assert.propContains(model.error, {
       legalName: { message: 'Vul de juridische naam in' },
       organizationStatus: { message: 'Selecteer een optie' },
     });
  });
 });
```

mirdono (changes requested):

app/models/registered-organization.js
wolfderechter: Got it, thanks!

mirdono (approved):
Looks good! :rocket:

PR #653, opened 2024-10-01
fix: ensure correct assignment of involved organizations for "has a relation with" memberships

mirdono:
## Problem

For "has a relation with" memberships the `updateMembershipRole` does not always
have enough information to correctly assign the current organization as either
`member` or `organization`. This is due to the fact that the 'correct'
assignment depends on what kind of other involved organization the user selects.

For example, if the current organization is a Worship Service it should be
assigned as
- `member` when the other organization is a Central Worship Service; or
- as `organization` when the other organization is a Representative Body.

The previous implementation, incorrectly, always assigned Worship Services as
`organization`, irrelevant of the classification of the other involved
organization.

## Proposed solution

The `updateMembershipRole` function now assigns the current organization as `member` or `organization` solely based on the membership role label. This was already the case for memberships with other roles than "has a relation with". For "has a relation with" memberships this means that the current organization will always be assigned as member, as the role label and inverse role label are identical.

The `save` function now performs an extra check and, if necessary, swaps the assignments for `member` and `organization`. This ensures that all such memberships are stored in a consistent manner.

### DIFF
--- a/app/constants/memberships.js
+++ b/app/constants/memberships.js
@@ -167,6 +167,32 @@ export const allowedHasRelationWithMemberships = [
   },
 ];
 
+/**
+ * Check whether the organization assignments in the given membership should be
+ * swapped in order become a valid "has a relation with" membership.
+ * The result of this function is only meaningful when applied to a membership
+ * - with the "has a relation with" role; and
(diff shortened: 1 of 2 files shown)


claire-lovisa (approved):
Works nicely :smile:

PR #654, opened 2024-10-07
fix: corrections to allowed "has a relation with" memberships

mirdono:
The allowed "has a relation with" memberships contained two errors which are fixed by this PR.

## Removed duplicate allowed membership between municipalities and provinces
Previously it was allowed for municipalities and provinces to be assigned as
either `organization` or `membner` in memberships concerning these two kinds of
organizations. More precisely, memberships with the following assignments were
both allowed:
| organization | member       |
|--------------|--------------|
| municipality | province     |
| province     | municipality |

Only the latter should be allowed. The former was removed as an allowed "has a
relation with" membership.

## Allow provinces to be related with both kinds of PEVAs
Previously a province could only be related to PEVA provinces. This is extended
to also include PEVA municipalities.

### DIFF
--- a/app/constants/memberships.js
+++ b/app/constants/memberships.js
@@ -135,7 +135,6 @@ export const allowedHasRelationWithMemberships = [
     organizations: [...MunicipalityCodeList],
     members: [
       ...OCMWCodeList,
-      ...ProvinceCodeList,
       ...DistrictCodeList,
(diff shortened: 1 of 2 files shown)


claire-lovisa (approved):
There is just a small complaint by the linter, but works fine from what I tested !

PR #660, opened 2024-11-21
Ember 5.12

Windvis:
This updates the app to Ember v5.12 so we can start preparing it for v6.

Builds on #659 so that needs to be merged first.


package-lock.json
Windvis, line 1: I rerolled the whole lock file since npm was messing up regular updates.

package.json
```diff
@@ -74,33 +78,33 @@
     "ember-cli-sass": "^11.0.1",
     "ember-concurrency": "^3.1.1",
     "ember-config-helper": "^0.1.3",
-    "ember-data": "~5.3.3",
+    "ember-data": "~5.3.8",
+    "ember-data-table": "^2.1.0",
```
Windvis, line 82: This is needed because Appuniversum no longer uses this addon but we do and we didn't list it as a dependency (which only worked by accident because Appuniversum brought it in).

PR #662, opened 2025-01-16
fix: determine filters on organisations page based on user's role

mirdono:
Previously the shown filters were based on whether the results in the table
contained worship organisations or not. This resulted in undesired behaviour
when there were no resulting organisations, with the shown filters falling back
to the non-worship ones.

This PR changes that behaviour to determine the shown filters based on the role
of the authenticated user. This results in consistently showing the correct
filters as a user's role remains the same within a single session.

## How to reproduce the initial bug
1. Log in as a worship user to OP
2. Navigate to the organisations overview page
3. Take note of the available filters on the left-hand side, especially those
   labelled *Type* and *Soort eredienst*.
3. Fill in one or more filters in the sidebar such that there are no resulting
   organisations. For example,
   - enter a bogus name in the *Naam* filter; or
   - select a municipality without worship organisations, such as *Aalter (voor
     fusie)*, in the *Gemeente* filter
4. A new *Type organisatie* filter should appear above the *Type* filter and the
   *Soort eredienst* filter should have disappeared.

## Related ticket
- OP-3520

### DIFF
--- a/app/controllers/organizations/index.js
+++ b/app/controllers/organizations/index.js
@@ -2,10 +2,11 @@ import Controller from '@ember/controller';
 import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
 import { tracked } from '@glimmer/tracking';
-import { CLASSIFICATION } from 'frontend-organization-portal/models/administrative-unit-classification-code';
 
 export default class OrganizationsIndexController extends Controller {
   @service router;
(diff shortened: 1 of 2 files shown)


Riadabd (approved):
Was able to reproduce the bug, and the PR fixes the issue as expected :+1:

PR #665, opened 2025-02-13
[OP-3386] Improve the local involvements percentage validation

Windvis:
The previous implementation validated the sum of percentages for each local involvement in the list. We now moved this logic to the worship-service model so it happens only once.

### DIFF
--- a/app/templates/organizations/organization/local-involvements/edit.hbs
+++ b/app/templates/organizations/organization/local-involvements/edit.hbs
@@ -129,7 +129,13 @@
           </td>
           {{#if @model.organization.isWorshipService}}
             <td>
-              {{#let involvement.error.percentage.message as |errorMessage|}}
+              {{#let
+                (or
+                  @model.organization.error.involvements.message
+                  involvement.error.percentage.message
+                )
+                as |errorMessage|
+              }}
                 <AuInput
                   @disabled={{this.isDisabledPercentage involvement}}
                   @error={{errorMessage}}
--- a/app/templates/organizations/organization/local-involvements/edit.hbs
+++ b/app/templates/organizations/organization/local-involvements/edit.hbs
@@ -131,8 +131,8 @@
             <td>
               {{#let
                 (or
-                  @model.organization.error.involvements.message
                   involvement.error.percentage.message
+                  @model.organization.error.involvements.message
                 )
                 as |errorMessage|
               }}
--- a/app/templates/organizations/organization/local-involvements/edit.hbs
+++ b/app/templates/organizations/organization/local-involvements/edit.hbs
@@ -130,24 +130,32 @@
           {{#if @model.organization.isWorshipService}}
             <td>
               {{#let
-                (or
-                  involvement.error.percentage.message
-                  @model.organization.error.involvements.message
-                )
-                as |errorMessage|
+                (this.isDisabledPercentage involvement)
+                as |isDisabledPercentage|
               }}
-                <AuInput
-                  @disabled={{this.isDisabledPercentage involvement}}
-                  @error={{errorMessage}}
-                  type="number"
-                  step="any"
-                  aria-labelledby="financial-percentage"
-                  value={{involvement.percentage}}
-                  {{on "input" (fn this.handlePercentageChange involvement)}}
-                />
-                {{#if errorMessage}}
-                  <AuHelpText @error={{true}}>{{errorMessage}}</AuHelpText>
-                {{/if}}
+                {{#let
+                  (and
+                    (not isDisabledPercentage)
+                    (or
+                      involvement.error.percentage.message
+                      @model.organization.error.involvements.message
+                    )
+                  )
+                  as |errorMessage|
+                }}
+                  <AuInput
+                    @disabled={{isDisabledPercentage}}
+                    @error={{errorMessage}}
+                    type="number"
+                    step="any"
+                    aria-labelledby="financial-percentage"
+                    value={{involvement.percentage}}
+                    {{on "input" (fn this.handlePercentageChange involvement)}}
+                  />
+                  {{#if errorMessage}}
+                    <AuHelpText @error={{true}}>{{errorMessage}}</AuHelpText>
+                  {{/if}}
+                {{/let}}
               {{/let}}
             </td>
             <td class="align-middle">
--- a/tests/unit/models/local-involvement-test.js
+++ b/tests/unit/models/local-involvement-test.js
@@ -66,7 +66,6 @@ module('Unit | Model | local involvement', function (hooks) {
 
     [
       ['0', 'Het percentage moet groter zijn dan 0'],
-      ['101', 'Het percentage mag niet groter zijn dan 100'],
       ['not a number', 'Vul het percentage in'],
       ['', 'Vul het percentage in'],
       [undefined, 'Vul het percentage in'],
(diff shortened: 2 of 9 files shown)


Windvis:
That took me longer than expected, `joi` isn't very intuitive to use so had to do a lot of reading / trial and error :smile:

app/templates/organizations/organization/local-involvements/edit.hbs
```diff
@@ -129,7 +129,13 @@
           </td>
           {{#if @model.organization.isWorshipService}}
             <td>
-              {{#let involvement.error.percentage.message as |errorMessage|}}
+              {{#let
+                (or
+                  @model.organization.error.involvements.message
+                  involvement.error.percentage.message
```
mirdono, line 135: **Nitpick:** I would be inclined to switch the order of these statements. This way the error messages concerning individual percentages not in the ]0,100] range are shown before any error message concerning the sum. For some reason that makes more sense me, and coincidentally it matches the previous behaviour.

tests/unit/models/local-involvement-test.js
```diff
@@ -66,7 +66,6 @@ module('Unit | Model | local involvement', function (hooks) {
 
     [
       ['0', 'Het percentage moet groter zijn dan 0'],
-      ['101', 'Het percentage mag niet groter zijn dan 100'],
```
mirdono, line 69: **Question:** Why did you remove this test? Since it concerns an individual percentage I think it is still a relevant test.

**Remark:** The stub code for other local involvements in this test can be removed. `External` validations are only executed if no other validations fail. Since these tests are expected to fail on the `then: Joi.number().min(1).max(100).required(),` validations the `external` validation will not be executed and the `getOtherLocalInvolvements` function is not called. Note, in retrospect this was already the case when I first added these tests so those lines were unnecessary from the start.

app/templates/organizations/organization/local-involvements/edit.hbs
```diff
@@ -129,7 +129,13 @@
...
+                )
+                as |errorMessage|
```
mirdono, line 137: **Issue:** The total percentage error is now also shown on disabled input fields, i.e. those for advisory local involvements. This seems confusing for a user. We probably need an additional check somewhere to not show percentage-related error messages for input fields where `isDisabledPercentage` is `true`.

mirdono (changes requested):
**Issue:** When a user cancels changes any already displayed errors are not reset. To reproduce:
1. Edit a local involvement table such that it would trigger an error, e.g. sum of percentages above 100
2. Try to save the faulty input so that the error messages are triggered
3. Cancel the made changes using the cancel button
4. On the local involvements view page click the edit button again
5. On the new edit page the previous errors are immediately displayed

mirdono (commented):
**Remark:** Strictly speaking outside the scope of the ticket, but should the validations concerning multiple supervisory local involvements not also be move to the worship administrative unit model? Same as the percentage sum validation this is an external validation for individual local involvements that will retrieve the other involvements anyway. Moving this to the model level likely helps us to get rid of the `getOtherLocalInvolvements` logic in the `local-involvement` model.

mirdono (commented):
**Issue:** Checks for financing percentages should probably move to the `worship-service` model level instead of `worship-administrative-unit` model. For central worship services, whose model is a subclass of the latter, financing percentages are not relevant.

tests/unit/models/local-involvement-test.js
Windvis: I think I was just overly zealous when removing things and misread what the test did :sweat_smile:

app/templates/organizations/organization/local-involvements/edit.hbs
Windvis: Yes, that makes a lot more sense :+1:

Windvis:
> **Remark:** Strictly speaking outside the scope of the ticket, but should the validations concerning multiple supervisory local involvements not also be move to the worship administrative unit model? Same as the percentage sum validation this is an external validation for individual local involvements that will retrieve the other involvements anyway. Moving this to the model level likely helps us to get rid of the `getOtherLocalInvolvements` logic in the `local-involvement` model.

I think the real issue is that the validations shouldn't exist on the model classes in the first place. It makes things a lot more complex than it should be and it will make updating EmberData to the newer patterns (requestManager) a lot more work, since those no longer use the Model class either.

An issue I ran into was that I had to load extra relationship data simply to make the validator happy, even though that data wasn't being edited in the form. We _could_ convert more things to conditional dependencies so it wouldn't be needed, but that's more work and gets complicated fast.

If we're going to spend more time on refactoring validations I would spend it on moving away from the hard coupling between the model class and validation schema and simply create a schema for each form instead.

You are right though, moving that logic elsewhere makes sense :smile: 

`</end-rant>` :sweat_smile:

Windvis:
(I think I updated most remarks, still have to move things to the worship-service class)

tests/unit/models/local-involvement-test.js
```diff
@@ -102,31 +96,7 @@ module('Unit | Model | local involvement', function (hooks) {
...
-      );
-      getOtherLocalInvolvementsStub.resolves([]);
-
-      const isValid = await model.validate();
-
-      assert.true(isValid);
-    });
-
-    test('it does not return an error when the sum of percentages is 100', async function (assert) {
+    test('it does not validate the sum of percentages', async function (assert) {
```
Windvis, line 99: @mirdono Do you think it's useful to keep this test?

mirdono: I go a bit back and forth on where to draw the line myself. While developing, especially when refactoring, I tend to write (too) many test in order to make sure I understand and capture the intended/existing behaviour. In the longer run probably not all of these tests are relevant enough to keep. But then I usually think to myself, I have already written these tests, and we can always remove them later if they really become obsolete. I am inclined to apply that same reasoning here. :-)

mirdono:
> > **Remark:** Strictly speaking outside the scope of the ticket, but should the validations concerning multiple supervisory local involvements not also be move to the worship administrative unit model? Same as the percentage sum validation this is an external validation for individual local involvements that will retrieve the other involvements anyway. Moving this to the model level likely helps us to get rid of the `getOtherLocalInvolvements` logic in the `local-involvement` model.
> 
> I think the real issue is that the validations shouldn't exist on the model classes in the first place. It makes things a lot more complex than it should be and it will make updating EmberData to the newer patterns (requestManager) a lot more work, since those no longer use the Model class either.
> 
> An issue I ran into was that I had to load extra relationship data simply to make the validator happy, even though that data wasn't being edited in the form. We _could_ convert more things to conditional validations so it wouldn't be needed, but that's more work and gets complicated fast.
> 
> If we're going to spend more time on refactoring validations I would spend it on moving away from the hard coupling between the model class and validation schema and simply create a schema for each form instead.
> 
> You are right though, moving that logic elsewhere makes sense :smile: 
> 
> `</end-rant>` :sweat_smile: 
> 

Personally, I think it makes more sense to have the validations in the corresponding model class as the model is (or at least should be) the information expert with respect to what it means for a model instance to be valid. But maybe the underlying discussion is what exactly the validations should be validating: model instances or forms. I am inclined to argue the former as models represent the domain concepts to which the business rules captured by the validations apply. At least this is how I tend to reason about these concepts. (But that might say more about me, than about what is the "good" way to implement this stuff.)

In any case, since our current approach seems to be incompatible with newer/future EmberData versions we can reconsider all of this when we update EmberData. :-s

mirdono (approved):
Thanks for addressing my comments. It all looks good now.

Windvis:
> Personally, I think it makes more sense to have the validations in the corresponding model class as the model is (or at least should be) the information expert with respect to what it means for a model instance to be valid. But maybe the underlying discussion is what exactly the validations should be validating: model instances or forms. I am inclined to argue the former as models represent the domain concepts to which the business rules captured by the validations apply. At least this is how I tend to reason about these concepts. (But that might say more about me, than about what is the "good" way to implement this stuff.)

I think this is turning into a "theory vs practice" discussion. Let's just agree to disagree.

PR #666, opened 2025-02-17
[OP-3539]: show kbo numbers for organizations in the dropdown

wolfderechter:
## ID
OP-3539

## Description

Show the KBO number for orgs in the org dropdown (as used in change events).

## How to test

1. Log in as `Josephine WorshipEditeerder`
2. Go to organisaties
3. Select a random org, go to `veranderingsgebeurtenissen` and create a new one
4. Choose `Wijziging gebiedsomgeving` for example
5. At the bottom the field `Betrokken organisatie` will appear, this should now show KBO numbers of the orgs.

### DIFF
--- a/app/models/organization.js
+++ b/app/models/organization.js
@@ -9,6 +9,7 @@ import {
   validateStringOptional,
 } from '../validators/schema';
 import getOppositeClassifications from '../constants/memberships';
+import { ID_NAME } from './identifier';
 
 export default class OrganizationModel extends AgentModel {
   @attr name;
@@ -127,11 +128,13 @@ export default class OrganizationModel extends AgentModel {
   }
 
   get kboNumber() {
-    const identifiers = this.identifiers.content;
+    const identifiers = this.hasMany('identifiers').value();
 
     for (const identifier of identifiers) {
-      const structuredIdentifier = identifier.get('structuredIdentifier');
-      if (identifier.idName === 'KBO nummer') {
+      const structuredIdentifier = identifier
+        .belongsTo('structuredIdentifier')
+        .value();
+      if (identifier.idName === ID_NAME.KBO) {
         return structuredIdentifier.get('localId');
       }
     }
--- a/app/models/organization.js
+++ b/app/models/organization.js
@@ -126,6 +126,19 @@ export default class OrganizationModel extends AgentModel {
     });
   }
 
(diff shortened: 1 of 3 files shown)


mirdono (approved):
Looks good

app/models/organization.js
```diff
@@ -126,6 +126,19 @@ export default class OrganizationModel extends AgentModel {
...
+
+    for (const identifier of identifiers) {
+      const structuredIdentifier = identifier.get('structuredIdentifier');
+      if (identifier.idName === 'KBO nummer') {
+        return structuredIdentifier.get('localId');
+      }
+    }
+
+    return null;
+  }
```
Windvis, line 140: `relationshipName.content` is a private property so I wouldn't use that. Did you find that documented somewhere?

If you want sync access to the loaded data of async relationships you can use the reference APIs:
- https://api.emberjs.com/ember-data/release/classes/HasManyReference
- https://api.emberjs.com/ember-data/release/classes/BelongsToReference

Which would look like this:

```suggestion
  get kboNumber() {
    const identifiers = this.hasMany('identifiers').value();

    for (const identifier of identifiers) {
      const structuredIdentifier = identifier.belongsTo('structuredIdentifier').value();
      if (identifier.idName === 'KBO nummer') {
        return structuredIdentifier.localId;
      }
    }

    return null;
  }
```
It also gets rid of the `.get` which is almost always a code smell :smile:

Windvis (changes requested):

app/models/organization.js
```diff
@@ -126,6 +126,19 @@ export default class OrganizationModel extends AgentModel {
     });
   }
 
+  get kboNumber() {
+    const identifiers = this.identifiers.content;
+
+    for (const identifier of identifiers) {
+      const structuredIdentifier = identifier.get('structuredIdentifier');
+      if (identifier.idName === 'KBO nummer') {
```
Windvis, line 134: It might also be nice to use the exported constant: https://github.com/lblod/frontend-organization-portal/blob/590bef4fdcc48eef433e566b7f7c5a59c75b517c/app/models/identifier.js#L6

app/models/organization.js
wolfderechter: Ahh I see, that's much better indeed. Thanks for the feedback

PR #667, opened 2025-02-18
chore: add release instructions to README

mirdono:
This ensures the release instructions are easy to find when new people join the
development team, or when someone needs to refresh their memory.

The entry is based on the one used in the Loket [frontend](https://github.com/lblod/frontend-loket/blob/development/README.md#releasing-a-new-version).

Thanks to @wolfderechter for pointing this out.


claire-lovisa (approved):
Thanks for the nice update!

PR #668, opened 2025-02-20
[OP-3550] Fix kbo numbers

wolfderechter:
## ID 

OP-3550

## Description

The kbo number implementation of https://github.com/lblod/frontend-organization-portal/pull/666 was missing an include which resulted in no kbo's being found. This PR also adds a check, so orgs with no KBO do not throw errors.

## Testing instructions

1. Go to frontend and log in as editor (non-worship)
2. Open your browser’s developer console
3. On the organisations page select municipality Aalst (others might work as well)
4. Go to the related organisations tab
5. Click the edit button in the top-right corner
6. The edit page should correctly load and the developer console should not contains errors

Best to check all places where there is a dropdown for orgs, but no errors should be thrown now


mirdono:
**Issue:** Ideally, the KBO number should also be displayed for the resulting organisation of a change event. When the involved organisations have the same name it is now not possible to differentiate them:
2025-02-20-161857_1630x249_scrot

Unfortunately, the relevant part of the template uses a plain`PowerSelect` instead of our customised `OrganizationSelect`: https://github.com/lblod/frontend-organization-portal/blob/8b437bb0637049b4a2a7dcbee8df080a0cd266d6/app/templates/organizations/organization/change-events/new.hbs#L306-L313

app/routes/organizations/organization/related-organizations/edit.js
```diff
@@ -28,6 +28,7 @@ export default class OrganizationsOrganizationRelatedOrganizationsEditRoute exte
 
     let query = {
       include: 'role,member,organization',
+
```
mirdono, line 31: **Nitpick:** This extra blank line is not needed.

mirdono (changes requested):
One issue, see separate comment.

mirdono (approved):
Looks good now, thanks for fixing this.

**Nitpick:** could you squash the commits a bit before merging? A commit and its reverting don't really need to be part of the main history.

PR #669, opened 2025-02-20
refactor: simplify condition to display classification label

mirdono:
A small simplification of some code that caught my eye reviewing another PR:
- Do not explicitly check whether an organisation has a classification, all
  organisations in OP have one.
- Use the appropriate model function to check whether an organisation is a
  worship administrative unit, instead of comparing classification labels.


wolfderechter (approved):
Looks much better indeed

PR #673, opened 2025-02-28
fix: correctly retrieve resulting organisations for new change event

mirdono:
Use the `hasMany` function to correctly resolve the async
`resultingOrganizations` relationship before accessing its elements. This
resolves a bug where creating a new merger change event results in errors being
thrown.

## How to reproduce the bugs
1. Log in as worship editor user
2. Select a (central) worship service from the organisations overview, any
   should do
3. Navigate to the change events tab (*nl. Veranderingsgebeurtenissen*)
4. Create a new change event by clicking the "+ Nieuw" button in the top-right
   corner
5. As change event type select a merger (*nl. Samenvoeging*)
6. Fill in at least the required fields
7. Try to save the change event by clicking the save button in the top-right
   corner. A `TypeError` should appear in the developer console.

For non-worship organisations the same bug can be triggered:
1. Log in as editor user
2. Select an IGS organisation (Projectvereniging, Dienstverlenende vereniging,
   Opdrachthoudende vereniging, or Opdrachthoudende vereniging met private
   deelname), any should do
3. Navigate to the change events tab (*nl. Veranderingsgebeurtenissen*)
4. Create a new change event by clicking the "+ Nieuw" button in the top-right
   corner
5. As change event type select a merger (*nl. Fusie*)
6. Fill in at least the required fields
7. Try to save the change event by clicking the save button in the top-right
   corner. A `TypeError` should appear in the developer console.

## Related ticket
- OP-3353

### DIFF
--- a/app/controllers/organizations/organization/change-events/new.js
+++ b/app/controllers/organizations/organization/change-events/new.js
@@ -180,9 +180,10 @@ export default class OrganizationsOrganizationChangeEventsNewController extends 
             if (currentOrganization.isCentralWorshipService) {
               resultingStatusId = ORGANIZATION_STATUS.INACTIVE;
             } else {
-              resultingStatusId = changeEvent.resultingOrganizations.includes(
-                organization,
-              )
+              const resultingOrganizations = changeEvent
+                .hasMany('resultingOrganizations')
+                .value();
+              resultingStatusId = resultingOrganizations.includes(organization)
                 ? ORGANIZATION_STATUS.ACTIVE
                 : ORGANIZATION_STATUS.INACTIVE;
             }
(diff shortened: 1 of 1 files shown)


app/controllers/organizations/organization/change-events/new.js
```diff
@@ -180,9 +180,10 @@ export default class OrganizationsOrganizationChangeEventsNewController extends
             if (currentOrganization.isCentralWorshipService) {
               resultingStatusId = ORGANIZATION_STATUS.INACTIVE;
             } else {
-              resultingStatusId = changeEvent.resultingOrganizations.includes(
-                organization,
-              )
+              const resultingOrganizations = changeEvent
+                .hasMany('resultingOrganizations')
```
Windvis, line 184: You could also simply await / yield the relationship here since it's in a place that can deal with the async'ness? This assumes the relationship data is always loaded before this code runs (which might be the case already, not familiar with the code).

mirdono: > You could also simply await / yield the relationship here since it's in a place that can deal with the async'ness? This assumes the relationship data is always loaded before this code runs (which might be the case already, not familiar with the code).

I *think* the necessary data is already loaded at this point, as it was filled in by the user in the form. Some testing does show that using `yield`, as used in other places in this function, also resolves the reported bug. Is there a technical reason reason to prefer one of over the other?

Windvis: In general, if you're in a place where you can handle the async I would await the relationship since that ensures everything is there, but if you're 100% sure the data is loaded it doesn't matter that much I think. 

Since the relationship data is added in this form it's probably fine. In other cases it might be safer to go for the await (or use the load/reload method on the hasMany reference before accessing it) to ensure everything is loaded in case we would accidentally remove an "include" somewhere.

Windvis (approved):

PR #674, opened 2025-02-28
refactor: merge both confirmation modals into a single one

mirdono:
This merges two confirmation modals, added independently, of each other into a single more generic modal. More specifically this concerns
- the modal introduced in #650 asking users to confirm when deleting a founder membership between organisations; and
- the modal asking users to confirm when they add a local involvement with a non-active municipality #670 or when they add a membership with a non active organisation #671

Furthermore, this more generally applies the feedback on the initial implementation in #671 to all modal instances:
- Put the modal outside of table sections to avoid an instances is created for each row.
- Apply Appuniversum's guidelines concerning buttons on modals:
  - right align the buttons on the modal
  - use the `secondary` skin for the cancel button
  - do not use `alert` for the confirm button


mirdono:
Closed, as superseded by #675

PR #675, opened 2025-03-04
refactor: merge confirmation modals into a single one

mirdono:
This merges two confirmation modals, added independently of each other, into a
single more generic modal. More specifically this concerns:
- the modal introduced in #650 asking users to confirm when deleting a founder
  membership between organisations; and
- the modal asking users to confirm when they add a local involvement with a
  non-active municipality as added in #670 or when they add a membership with a
  non active organisation which was added in #671

This also applies the feedback on the initial implementation in #671 to all modals:
- Put the modal for confirming deleting founders outside of table sections to
  avoid that a modal instance is created for each row. This was noticeable by
  the background turning black.
- Apply Appuniversum's guidelines concerning buttons on to all modals:
  - right align the buttons on the modal
  - use the `secondary` skin for the cancel button
  - do not use `alert` for the confirm button


app/controllers/organizations/organization/related-organizations/edit.js
```diff
@@ -148,6 +154,12 @@ export default class OrganizationsOrganizationRelatedOrganizationsEditController
     return membership.getRoleLabelForPerspective(this.model.organization);
   }
 
+  @action
+  getConfirmationMessageForFounderDeletion() {
```
Windvis, line 158: I think this can be an actual getter? At the moment this method is invoked as a helper in the template, which seems unneeded?

app/components/confirmation-modal.hbs
```diff
@@ -1,11 +1,9 @@
-{{! TODO: This model should be renamed from "inactive" to "non-active". Or it should be merged with the other confirmation modal into a single general confirmation modal.}}
 <AuModal @modalOpen={{true}} @closeModal={{@onCancel}}>
   <:title>Bevestig</:title>
-  <:body>Opgelet, je hebt een niet actieve organisatie geselecteerd. Ben je
-    zeker dat je deze relatie wenst toe te voegen?
-  </:body>
+  <:body>{{@message}}</:body>
   <:footer>
-    <AuToolbar @reverse={{true}} as |Group|>
```
Windvis, line 8: Didn't this work as expected or something?

app/helpers/confirmation-message-text.js
```diff
@@ -0,0 +1,15 @@
+import { helper } from '@ember/component/helper';
+import { assert } from '@ember/debug';
+
+const CONFIRMATION_MESSAGE_TEXT = {
+  nonActiveRelatedOrganization:
+    'Opgelet, je hebt een niet actieve organisatie geselecteerd. Ben je zeker dat je deze relatie wenst toe te voegen?',
+};
+
+export default helper(function confirmationMessageText([key]) {
```
Windvis, line 9: Wrapping functions in the helper util is no longer needed since ember-source 4.5+ (the blueprint was simply outdated for a very long time).

```suggestion
export default function confirmationMessageText(key) {
```

Windvis (commented):
Some small remarks, but looks good to go otherwise.

app/controllers/organizations/organization/related-organizations/edit.js
mirdono: > I think this can be an actual getter? At the moment this method is invoked as a helper in the template, which seems unneeded?

You are right. Initially I had written this function differently, involving an argument. Forgot to change it to a getter when the implementation changed.

app/helpers/confirmation-message-text.js
mirdono: > Wrapping functions in the helper util is no longer needed since ember-source 4.5+ (the blueprint was simply outdated for a very long time).
> 
> ```suggestion
> export default function confirmationMessageText(key) {
> ```

Thanks for point that out. Fixed it for this instance. Will probably take a stab at doing the same for other helpers at some point.

app/components/confirmation-modal.hbs
Windvis: the `@reverse` argument was removed and replaced again with an empty `<Group />` so I assumed things maybe didn't work as expected (and there would be a bug in Appuniversum).

Windvis: Wait, now it's fine :smile: maybe I was looking at an outdated page when I posted that, sorry for the noise.

Windvis (approved):

Windvis:
Very strange indeed. It mentions a force push to an older commit, not sure what happened there.

PR #677, opened 2025-03-25
[OP-3563] Human readable phone numbers

Windvis:
Based on the [`tel-format` helper](https://github.com/lblod/frontend-contactgegevens-loket/blob/78ab121f0e2807695f01e6edaed6dc23dda3a2fe/app/helpers/tel-format.js) in contactgegevens-loket.

### DIFF
--- a/app/models/contact-point.js
+++ b/app/models/contact-point.js
@@ -16,7 +16,7 @@ export const CONTACT_TYPE = {
 
 export default class ContactPointModel extends AbstractValidationModel {
   @attr email;
-  @attr telephone;
+  @attr('phone') telephone;
   @attr fax;
   @attr website;
   @attr type;
--- a/tests/unit/helpers/tel-format-test.js
+++ b/tests/unit/helpers/tel-format-test.js
@@ -1,0 +1,98 @@
+import telFormat from 'frontend-organization-portal/helpers/tel-format';
+import { module, test } from 'qunit';
+
+module('Unit | Helper | tel-format', function () {
+  test('it formats phone numbers into a human readable format', function (assert) {
+    assert.strictEqual(
+      telFormat('+3221524562'),
+      '+32 21 52 45 62',
+      'it supports phone numbers with a + country prefix',
+    );
+    assert.strictEqual(
+      telFormat('+32482455642'),
+      '+32 482 45 56 42',
+      'it supports phone numbers with a + country prefix',
+    );
+    assert.strictEqual(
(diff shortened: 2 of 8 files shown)


tests/unit/helpers/tel-format-test.js
Windvis, line 1: We should port these tests back to the contactgegevens-loket app since they don't have any tests for their version yet.

tests/unit/helpers/tel-format-test.js
```diff
@@ -0,0 +1,98 @@
...
+      '012 345',
+      'it returns irregularly short phone numbers as-is',
+    );
+
+    assert.strictEqual(
+      telFormat('012 345 67 81 23 28'),
+      '012 345 67 81 23 28',
+      'it returns irregularly long phone numbers as-is',
+    );
+  });
```
Windvis, line 81: This part is still waiting for feedback from Sofie. I think we can probably cover some more cases under the existing rules, even if they don't have the traditional length. Still unsure if that actually happens in our apps though.

app/models/contact-point.js
```diff
@@ -16,7 +16,7 @@ export const CONTACT_TYPE = {
 
 export default class ContactPointModel extends AbstractValidationModel {
   @attr email;
-  @attr telephone;
+  @attr('phone') telephone;
```
Windvis, line 19: I think this also means we can simply remove the [`transformPhoneNumbers` util](https://github.com/lblod/frontend-organization-portal/blob/0fc611807a35ff18e07f4188d383801470cad4a1/app/utils/transform-phone-numbers.js)?

Windvis: Yup, seems like it:

image

It doesn't cause issues though. The backend properly handles this scenario and only keeps a single prefix.

tests/unit/helpers/tel-format-test.js
Windvis: The current solution where we display the data as-is is fine for now :+1:

mirdono (approved):
Looks good, seems certainly more robust that what we had before. Thanks for also including a nice amount of tests, makes it easier to understand the different cases.

Windvis:
The code could use some more refactoring, but I'll deal with that when I move things to Appuniversum, unless there are some glaring issues I missed.

cecemel (approved):
Looks good. Code is very easy to read. Testing around yields nothing unexpected.

PR #678, opened 2025-03-31
chore: update Woodpecker CI configuration

mirdono:
- Update the `ember-cli` in the PR test CI to match the version used
- RPIO's Woodpecker CI was updated to v3


claire-lovisa (commented):
Might be worth adding a `.chore.yml` file for the chore branch prefix? We use it quite often and it could be handy to check if the CI works well from this branch?

mirdono:
> Might be worth adding a `.chore.yml` file for the chore branch prefix? We use it quite often and it could be handy to check if the CI works well from this branch?

I added a `.chore.yml` for testing purposes, but I am hesitant to keep it permanently. While it is useful if we want to deploy a specific branch for some reason, I think that most of the time the resulting image is not actively used. It seems like a waste of resources in that case and it pollutes the docker hub with unnecessary images. Arguably, the same goes for the `.feature.yml` and `.refactor.yml`. In the past we have had good reason to use them, but I think most of these branches are simply approved and merged into development.

I would be inclined to further rework our CI configurations. Top of my head maybe something like:
- remove the configurations for refactor and chore branches
- change the feature branch configuration from a `push` event to `pull_request` event

I suspect more advanced configurations are possible to build only images we actually use (or approximate that situation better). But I am not familiar enough with Woodpecker to know for certain what is (not) possible.

claire-lovisa (commented):
@mirdono you're completely right, it's just using too much resources for almost nothing, sorry for the bad idea. At least now we saw it was working with the temporary chore file, let's remove it.
I also like your idea of the pull request event for feature branches, would probably cause a bit less unnecessary builds.

PR #679, opened 2025-06-03
[OP-3611]: remove date overlap validations

wolfderechter:
## ID

OP-3611

## Description

Remove the date overlap validations for governing bodies

## How to test

1. Go to an organizations 'Bestuursorgaan' and click edit
2. Fill in something like '31-03-2026' (fill in the value somewhere else if needed to trigger the overlap issue) for the end date which normally throws a 'geen overlap' error, it should now not throw that error anymore.


mirdono (approved):
I removed some additional helper functions and tests that now became obsolete. Otherwise, the changes were correct.

PR #680, opened 2025-06-11
feat: consult and edit an administrative unit's scope of operation

mirdono:
OP is extended with functionality to allow users to consult and edit the scope of operation (*nl. werkingsgebied*) for an administrative unit. This PR introduces the necessary frontend changes to use a newly introduced backend service for this functionality.

### Consult scope of operation
A new field is added to the core data page for organisations to display the organisation's scope of operation. The expected value for this field is a list of all municipality-level locations that are part of the organisation's scope of operation. (Yes this can become very long, just try opening a province :worried:)

### Edit scope of operation
The form to edit the core data for an organisation should allow users to select one or more municipality-level locations as its scope of operation. Similarly, the new organisation form should allow users to select appropriate locations when creating a new administrative unit.

The scope of operation is mandatory for all administrative units and an error should be shown if the users fails to select at least one value.

For municipalities and provinces their scope of operation cannot be edited.

## Proposed solution
This PR introduces a new service in the frontend that acts as a wrapper around the new `scope-of-operation-service` introduced in the backend. This wrapper service essentially translates between the frontend's models and the UUIDs required/returned by the backend service. Putting this logic within a single service allows to reuse it within different routes, such as the edit or new organisation route, without having to duplicate this logic.

Furthermore, a new component is introduced that allows users to select multiple municipality-level locations. This component is similar to the already existing components to select multiple organisations. This new component is added to the edit core data and new organisation form.

## Determine region for organisations
For some kinds organisations (municipalities, IGSs, PEVAs, a OCMW associations) their core data page contains an optional region field. The value of this field is also based on the location that is assigned as the organisation' scope of operation. More specifically, for municipalities it is the reference region in which its corresponding location is located. For the other kinds of organisations it is the reference region of the municipality in which their primary site is located.

The existing functionality for this mapping broke due to a required model change in `LocationModel` to allow that a location is contained within multiple other locations. Consequently, it was no longer possible to simply follow the `locatedWithin` relation as this changed to a `hasMany`. Therefore, this functionality was also revised to search for the reference region in which the municipality is located. This relies on the fact a municipality is at most in 1 reference regions, thus that there is no overlap between reference regions.

## Related PRs:
- Backend: [#550](https://github.com/lblod/app-organization-portal/pull/550)
- Service: [#1](https://github.com/lblod/scope-of-operation-service/pull/1)

## Related tickets:
- OP-3203
- OP-3600

### DIFF
--- a/app/components/location-multiple-select.hbs
+++ b/app/components/location-multiple-select.hbs
@@ -1,0 +1,19 @@
+<div class={{if @error "ember-power-select--error"}}>
+  <PowerSelectMultiple
+    @loadingMessage="Aan het laden..."
+    @noMatchesMessage="Geen resultaten"
+    @searchMessage="Typ om te zoeken"
+    @disabled={{@disabled}}
+    @allowClear={{true}}
+    @searchEnabled={{true}}
+    @searchField="label"
+    @search={{perform this.loadLocationsMultipleTask}}
+    @closeOnSelect={{false}}
+    @selected={{@selected}}
+    @onChange={{@onChange}}
+    @triggerId={{@id}}
+    as |location|
+  >
+    {{location.label}}
+  </PowerSelectMultiple>
+</div>
--- a/app/components/location-multiple-select.js
+++ b/app/components/location-multiple-select.js
@@ -1,0 +1,45 @@
+import Component from '@glimmer/component';
+import { inject as service } from '@ember/service';
+import { restartableTask } from 'ember-concurrency';
+
+// TODO: open dropdown with all results on focus/open
+// TODO: clear input after selection
+export default class LocationMultipleSelectComponent extends Component {
+  @service store;
+
+  @restartableTask
+  *loadLocationsMultipleTask(searchParams = '') {
+    const provinces = this.args.provinceLocations;
+
+    const query = {
+      filter: { level: 'Gemeente' },
+      sort: 'label',
+      include: 'located-within',
+      // NOTE (21/05/2025): Make sure to load all municipality locations
+      page: { size: 400 },
+    };
+
+    if (searchParams.trim() !== '') {
+      query['filter[label]'] = searchParams;
+    }
+
+    const municipalities = yield this.store.query('location', query);
+
+    return this.extractProvinceGroups(provinces, municipalities);
+  }
+
+  extractProvinceGroups(provinces, municipalities) {
+    return provinces
+      .map((province) => this.createGroupForProvince(province, municipalities))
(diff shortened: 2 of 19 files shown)


claire-lovisa (changes requested):
I did find some strange things while testing, putting them below:
- I'm on the edit core data page of Aalst http://localhost:4200/organisaties/974816591f269bb7d74aa1720922651529f3d3b2a787f5c60b73e5a0384950a4/kerngegevens/edit , try to go back to the main page via the breadcrumb "Organisatie", error in the console `route-not-found`. Doesn't seem to happen on the dev branch.
- Still when on the edit page, I can't switch mock login account, the same error gets raised
- Is there a plan to add locations to worship services? The ones I checked didn't have locations and it blocks the update of those worship services, as it's mandatory to be able to save

mirdono:
> I did find some strange things while testing, putting them below:
> - I'm on the edit core data page of Aalst http://localhost:4200/organisaties/974816591f269bb7d74aa1720922651529f3d3b2a787f5c60b73e5a0384950a4/kerngegevens/edit , try to go back to the main page via the breadcrumb "Organisatie", error in the console `route-not-found`. Doesn't seem to happen on the dev branch.
> - Still when on the edit page, I can't switch mock login account, the same error gets raised

I am unable to replicate this behaviour in my setup. I did indeed ran into this error occasionally while developing/testing but could never pin down the cause or replicate it reliably. Does the error happen consistently for you in these situations?

> - Is there a plan to add locations to worship services? The ones I checked didn't have locations and it blocks the update of those worship services, as it's mandatory to be able to save

Business initially said it was mandatory for all administrative units. But it seems indeed strange for the case of (central) worship services. I asked for clarification on this in the ticket.

mirdono:
Aha, when I open the edit core data page for a province and try to leave by cancelling or via the breadcrumb I do reliably get errors. More specifically: `Error while processing route: route-not-found _ember_data_store_private__WEBPACK_IMPORTED_MODULE_5__.peekCache(...) is null`, and a `TypeError` essentially saying the same thing. Is this the error you also encounter?

claire-lovisa:
> Aha, when I open the edit core data page for a province and try to leave by cancelling or via the breadcrumb I do reliably get errors. More specifically: `Error while processing route: route-not-found _ember_data_store_private__WEBPACK_IMPORTED_MODULE_5__.peekCache(...) is null`, and a `TypeError` essentially saying the same thing. Is this the error you also encounter?

Yes it's this exact one!!

claire-lovisa (approved):
Good catch :sweat_smile:

PR #685, opened 2025-07-09
[OP-3601] Improve the phone input experience

Windvis:
- Allow entry of phone numbers without a `+`, so special numbers can be entered
- display warnings when users enter unusually short, unusually long or numbers not in the international notation.


wolfderechter (approved):
Seems to work as expected

PR #689, opened 2025-08-12
[OP-3655] don't clear the address when postcode value is unchanged

wolfderechter:
## ID

OP-3655

## Description

When tabbing through the contact edit card, an unchanged postal code will still trigger the clearing of municipality and province, even if it's not changed. So added a small check.

Since the postcode is inside a `TrimInput` component, this seemed like the easiest option to fix the bug since the `{{on "focusout" (perform this.trimInput)}}` will get triggered.

### DIFF
--- a/app/components/trim-input.js
+++ b/app/components/trim-input.js
@@ -4,7 +4,7 @@ import { restartableTask } from 'ember-concurrency';
 export default class TrimInputComponent extends Component {
   @restartableTask
   *trimInput(event) {
-    const input = yield event.target.value;
-    if (this.args.value !== input) this.args.onUpdate(input.trim());
+    const input = (yield event.target.value).trim();
+    if (this.args.value !== input) this.args.onUpdate(input);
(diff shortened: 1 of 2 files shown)


app/components/trim-input.js
```diff
@@ -5,6 +5,6 @@ export default class TrimInputComponent extends Component {
   @restartableTask
   *trimInput(event) {
     const input = yield event.target.value;
-    this.args.onUpdate(input.trim());
+    if (this.args.value !== input) this.args.onUpdate(input.trim());
```
Windvis, line 8: I think we need to compare the value with the trimmed input, no? Otherwise we get "false positive" updates when the user only added whitespace.

wolfderechter: For the tab itself it doesn't matter, but might be better to avoid the false positive whitespaces indeed

Windvis (approved):

PR #690, opened 2025-10-21
OP-3668: Frontend - Improvements on creating new organisation in Brussel municipality 

kikef98:
Frontend changes for the tickets under OP-3668. Changed the message for when no adress found and extended the werkingsgebieden selector and added a filter for the source so that the municipalities aren´t shown in the main organizations page. 

### Tickets
User Story:
- OP-3668

Specific tickets:
- OP-3671
- OP-3672
- OP-3673


claire-lovisa (approved):
Works perfectly! 😃

PR #702, opened 2026-02-18
OP-3735 Changed KBO number to test the change, now i can't change it back to the right one

kikef98:
**Problem:** In QA a KBO number was changed to test a change and then they were unable to change it back, since they got the message "Dit KBO nummer is al in gebruik". The reason for this was that the linked KboOrganisation still had this KBO number and the validation gave an error on that.

**Fix:** Edit to the validation so that it allows the same KBO Number if it's only on a type 'kbo-organization'. It still gives an error if its on a administrative-unit

**Test:** Change a KBO number to not the right one and see if you can change it back. Also good to test if you cannot change it to an already in use KBO number. (I also tested this).

### Ticket
OP-3735

### DIFF
--- a/app/models/identifier.js
+++ b/app/models/identifier.js
@@ -37,7 +37,7 @@ export default class IdentifierModel extends AbstractValidationModel {
             if (!localId.match(/^\d{10}$/)) {
               return helpers.message('Vul het (tiencijferige) KBO nummer in.');
             }
-            // KBO must be unique (valid only when changed)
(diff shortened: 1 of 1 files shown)


DamonKennes (approved):
- Tried changing the kbo number from Tielt-Winge the same way as in the ticket. Was able to change it and change it back without any errors.
- When changing the kbo number to an existing one (e.g. Antwerpen) it still shows the error as expected.

PR #703, opened 2026-03-05
fixed typo in a variable

bdevloed:
Bumped into this when looking into OP-3742 (does not explain/solve the issue though)


bdevloed:
Created a ticket for this: [OP-3750]

DamonKennes (approved):
Seems like it was indeed a typo. I checked in the browser and it gets set to null correctly with the change.

PR #704, opened 2026-05-21
[OP-3804] Add a "einddatum" column to the bestuursorgaan details page

Windvis:


app/templates/organizations/organization/governing-bodies/governing-body/index.hbs
```diff
@@ -194,10 +198,11 @@
               @currentSorting={{this.sort}}
               @label="Start mandaat"
             />
+            <th>Einde mandaat</th>
```
cecemel: @Windvis should it not be 'Geplande einddatum'?

Windvis: Maybe? The ticket wasn't clear about that. I assumed the current column should stay the same so I didn't change that part.

But I can change it.

So `Geplande einddatum` and `Einddatum`? Or something else?

cecemel (approved):

PR #706, opened 2026-06-10
[OP-3814] Only display the "geplande einddatum" column for worship organizations

Windvis:
It's not relevant for other organization types.

Follow-up to #704


DamonKennes (changes requested):
Geplande einddatum is now hidden for non-worship organizations in the bestuursleden table. It is still shown for the mandatarissen table below it.

Is this intended? If not, this also has to be hidden for non-worship organizations.

Windvis:
Whoops, my bad. I just missed that :grimacing: Thanks!

(I force pushed, so you need to reset your branch if you want to test)

DamonKennes (approved):

PR #707, opened 2026-06-17
Added juridische vorm field + change-event

DamonKennes:
- Added Juridische vorm [OP-3816]
- Added change-event to change juridische vorm [OP-3820]

Backend PR: https://github.com/lblod/app-organization-portal/pull/603


app/components/legal-form-select.js
```diff
@@ -0,0 +1,34 @@
...
+  findLegalFormById(id) {
+    if (this.loadLegalFormsTask.isRunning) {
+      return null;
+    }
+
+    let legalForms = this.loadLegalFormsTask.last.value;
+    return legalForms.find((legalForm) => legalForm.id === id);
+  }
+
+  @task *loadLegalFormsTask() {
```
Windvis, line 33: FYI, for future reference (no need to change in this PR), I would use the function version. The generator syntax is deprecated and support for it is removed in future ember-concurrency releases. (there is a codemod, so converting it is easy, but I wouldn't use this syntax for new tasks in the future.

```js
loadLegalFormsTask = task(async () => { await ... })
```

app/components/legal-form-select.js
```diff
@@ -0,0 +1,34 @@
...
+    return yield this.store.findAll('legal-form-code');
```
Windvis: This doesn't return all the options at the moment. mu-cl-resources paginates responses by 20 items by default, so 5 are missing. Since we know the number of items that currently exist, and they are unlikely to change often I would just increase the page size. You'll need to use `.query` instead of `.findAll` though.

```suggestion
    return yield this.store.query('legal-form-code', { 'page[size]': 30 });
```

app/components/legal-form-select.js
```diff
@@ -0,0 +1,34 @@
+import Component from '@glimmer/component';
+import { inject as service } from '@ember/service';
```
Windvis: Nitpick: FYI, this is the non-deprecated import.
```suggestion
import { service } from '@ember/service';
```

app/templates/organizations/organization/change-events/new.hbs
```diff
@@ -71,6 +71,23 @@
                 />
               </:content>
             </Item>
+            {{#if this.isLegalFormChange}}
```
Windvis, line 74: Not super familiar with the change event flows, but it seems there is also an edit page for change events, do we also want to display this field there (read only?). Nothing is shown at the moment, which seems a bit strange to me, but maybe that's how it's supposed to work?

Windvis (changes requested):
Mostly seems to work as expected 👍 

1 issue and some nitpicks / remarks.

app/templates/organizations/organization/change-events/new.hbs
DamonKennes: Good catch! I forgot that there was an edit page for the change-events. I'll add the field there read-only so the user have a better overview.

Windvis (approved):

app/models/concept-scheme.js
```diff
@@ -0,0 +1,5 @@
+import Model, { attr } from '@ember-data/model';
+
+export default class ConceptSchemeModel extends Model {
```
Windvis: I wouldn't add this unless we actually need it in the frontend.

app/models/concept.js
```diff
@@ -1,7 +1,14 @@
...
   @attr label;
   @attr altLabel;
   @attr notation;
+  @attr('number') order;
+
+  @belongsTo('concept-scheme', {
+    inverse: null,
+    async: true,
+  })
+  inScheme;
```
Windvis: Same for this, actually. We don't use these in our code since we construct the query manually. Adding them has little value other than documentation maybe, but would rather look at the resource config anyways.

app/models/concept.js
```diff
@@ -1,7 +1,14 @@
-import Model, { attr } from '@ember-data/model';
+import Model, { attr, belongsTo } from '@ember-data/model';
 
 export default class ConceptsModel extends Model {
   @attr label;
   @attr altLabel;
   @attr notation;
+  @attr('number') order;
```
Windvis: Oh, another FYI, the ` @attr('number')` number part is not needed since the backend already sends these as numbers. This isn't a "type" but a "transform" so it would only be needed if we received a string but wanted to use it as a number in our frontend code.

But anyway, since we don't read the order anywhere I would also not add it to the class. I try to keep things as simple as possible instead of trying to fully match the backend config for things we don't use. Less code to maintain, blabla 😅.

Windvis (commented):
Some nitpicks, but nothing blocking.

Windvis (approved):

PR #709, opened 2026-06-24
[OP-3818] Site type changes

Windvis:
Depends on this PR: https://github.com/lblod/app-organization-portal/pull/604

Several improvements
- Fix an issue with the site selector when reloading the page
- Display the `Andere vestiging` option for all organization types
- Add a new "Bezoekadres" option


DamonKennes (approved):
Looks like it works!

PR #710, opened 2026-06-26
[OP-3794] Add new organization types

bdevloed:
@Windvis cf https://github.com/lblod/app-organization-portal/pull/605

I've drafted some front end changes
- classifications have an altlabel (vervoerregio is used instead of vervoerregioraad )
- optional KBO - ILV and vervoerregioraad are not formal orgs

werkingsgebied for non bestuursorganen is different at RDF level, but mapped to the same scope in the resource config

### DIFF
--- a/app/components/classification-select.hbs
+++ b/app/components/classification-select.hbs
@@ -4,13 +4,13 @@
     @loadingMessage="Aan het laden..."
     @noMatchesMessage="Geen resultaten"
     @searchMessage="Typ om te zoeken"
-    @searchField="label"
+    @searchField="searchLabel"
     @options={{this.classifications.value}}
     @selected={{@selected}}
     @onChange={{@onChange}}
     @triggerId={{@id}}
     as |classification|
   >
-    {{classification.label}}
+    {{classification.label}}{{#if classification.altLabel}} ({{classification.altLabel}}){{/if}}
   </PowerSelect>
 </div>
--- a/app/controllers/organizations/new.js
+++ b/app/controllers/organizations/new.js
@@ -35,13 +35,19 @@ export default class OrganizationsNewController extends Controller {
 
   @tracked locations;
 
+  // ILV and Vervoerregioraad have no KBO number, so KBO is optional for them.
+  get requiresKbo() {
+    const model = this.currentOrganizationModel;
+    return !(model?.isInterlokaleVereniging || model?.isVervoerregioraad);
+  }
+
   get hasValidationErrors() {
     return (
       this.currentOrganizationModel.error ||
       this.model.address.error ||
       this.model.contact.error ||
       this.model.secondaryContact.error ||
-      this.model.identifierKBO.error ||
+      (this.requiresKbo && this.model.identifierKBO.error) ||
       this.model.identifierSharepoint.error ||
       this.memberships.some((membership) => membership.error) ||
       this.membershipsOfOrganizations.some((membership) => membership.error)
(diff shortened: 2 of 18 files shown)


Windvis:
This is still missing some of the requested text changes.

```
Textchange: Type ‘Vereniging’ > 'Vereniging algemeen'
Textchange: Type ‘Vennootschap’ > 'Vennootschap algemeen'
```

Not sure if this actually needed and if it how we best implement it. The labels come from the DB, but changing them there might not be correct if it's simply a presentation thing? We could hardcode some overrides in the frontend?

app/controllers/organizations/new.js
```diff
@@ -605,9 +616,13 @@ export default class OrganizationsNewController extends Controller {
...
-      structuredIdentifierKBO = setEmptyStringsToNull(structuredIdentifierKBO);
-      yield structuredIdentifierKBO.save();
-      yield identifierKBO.save();
+      if (this.requiresKbo) {
+        structuredIdentifierKBO = setEmptyStringsToNull(
+          structuredIdentifierKBO,
+        );
+        yield structuredIdentifierKBO.save();
+        yield identifierKBO.save();
+      }
```
Windvis: @bdevloed Should we hide the field if it's not required? It seems we're not saving it if it's not (so displaying it seems unnecessary)?

Windvis:
@bdevloed I did some small changes, could you review them?

Windvis:
Some more changes are needed after discussing with Sofie:

- We need to hide the "gerelateerde organisaties" section when creating one of the new types. The current fallback isn't correct. We should hide it behind a flag so we never unintentionally display this when new types are added. https://github.com/lblod/frontend-organization-portal/blob/6095058053634a0067a0c0b45f701427db37832b/app/templates/organizations/new.hbs#L494
- KBO number needs to be hidden instead of optional for the ones that don't have one also on the edit page
- ~~on the overview page we should hide it if it's not set (this might change in the future, but it would be consistent with other fields with optional data).~~ Already the case.

app/templates/organizations/new.hbs
```diff
@@ -334,8 +350,17 @@
...
+          this.currentOrganizationModel.isAgb
+          this.currentOrganizationModel.isDistrict
+          this.currentOrganizationModel.isIgs
+          this.currentOrganizationModel.isPoliceZone
+          this.currentOrganizationModel.isAssistanceZone
+          this.currentOrganizationModel.isOcmwAssociation
+          this.currentOrganizationModel.isPevaMunicipality
+          this.currentOrganizationModel.isPevaProvince
+          this.currentOrganizationModel.isApb
+        )
```
Windvis, line 363: One thing I noticed after this change (and the `isApb` check below is that `Vlaamse gemeenschapscommissie` no longer displays the related org section.

It was displaying the same case as the APB "fallback" so either that was on accident, or the comment was outdated and we actually want it there.

@bdevloed Do you know, or know how I can check? I'm missing the needed background context. I can also ask Sofie if needed.

bdevloed: Checked PRD: VGC ("Vlaamse Gemeenschapscommissie") has zero org:memberships in either direction. So the related-org section was never used for the VGC - it only fell into the APB fallback by accident, and would have required picking a related province, which makes no sense for the VGC. Hiding it in the allowlist is the correct behaviour.

bdevloed:
> This is still missing some of the requested text changes.
> 
> ```
> Textchange: Type ‘Vereniging’ > 'Vereniging algemeen'
> Textchange: Type ‘Vennootschap’ > 'Vennootschap algemeen'
> ```
> 
> Not sure if this actually needed and if it how we best implement it. The labels come from the DB, but changing them there might not be correct if it's simply a presentation thing? We could hardcode some overrides in the frontend?

I have added a migration to address this.

bdevloed:
> @bdevloed I did some small changes, could you review them?

Thanks, looks good to me.

app/components/classification-select.hbs
```diff
@@ -4,13 +4,13 @@
     @loadingMessage="Aan het laden..."
     @noMatchesMessage="Geen resultaten"
     @searchMessage="Typ om te zoeken"
-    @searchField="label"
+    @searchField="searchLabel"
     @options={{this.classifications.value}}
```
Windvis, line 8: One issue I missed is that this component fetches the data from the backend, but doesn't deal with pagination. So we only retrieve the first 20 results, and we have 23 now apparently. We need to increase the page size as a quick fix.

Windvis: PR: https://github.com/lblod/frontend-organization-portal/pull/716

PR #711, opened 2026-06-29
OP-3816: juridische vorm mandatory + editable

DamonKennes:
- juridische vorm mandatory + editable (Sofie comments on OP-3816)

### DIFF
--- a/app/templates/organizations/organization/core-data/edit.hbs
+++ b/app/templates/organizations/organization/core-data/edit.hbs
@@ -114,18 +114,18 @@
                 {{#unless @model.organization.isWorshipAdministrativeUnit}}
                   <Item
                     @labelFor="legal-form"
+                    @required={{true}}
+                    @errorMessage={{@model.organization.error.legalForm.message}}
                   >
                     <:label>Juridische vorm</:label>
-                    <:content>
-                      <PowerSelect
-                        @disabled={{true}}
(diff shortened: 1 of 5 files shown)


app/templates/organizations/organization/core-data/edit.hbs
```diff
@@ -114,18 +114,18 @@
                 {{#unless @model.organization.isWorshipAdministrativeUnit}}
                   <Item
                     @labelFor="legal-form"
+                    @required={{true}}
+                    @errorMessage={{@model.organization.error.legalForm.message}}
                   >
                     <:label>Juridische vorm</:label>
-                    <:content>
-                      <PowerSelect
-                        @disabled={{true}}
```
Windvis, line 121: Maybe we should only make it editable if there is no legalForm set yet? That way they can still fill it in, without allowing edits that require a change event? Or we only make it always editable if the user is an admin of some sort (not sure if OP has that).

Or we could put it behind a feature flag which we can then flip to disable editing without having to create a new build.

Windvis (approved):
Works as expected and is what Sofie requested. My other comment is just a potential improvement, but not needed.

app/templates/organizations/organization/core-data/edit.hbs
DamonKennes: I added it as a feature flag.

PR #713, opened 2026-07-10
[OP-3812]: Operation area filter

NikiDeVisscher:
This PR adds a werkingsgebied filter in the organization overview. 
Backend PR (needed for testing): https://github.com/lblod/app-organization-portal/pull/607

### DIFF
--- a/app/components/operation-area-select-by-name.js
+++ b/app/components/operation-area-select-by-name.js
@@ -1,0 +1,27 @@
+import Component from '@glimmer/component';
+import { inject as service } from '@ember/service';
+import { task } from 'ember-concurrency';
+import { trackedTask } from 'ember-resources/util/ember-concurrency';
+
+export default class OperationAreaSelectByNameComponent extends Component {
+  @service store;
+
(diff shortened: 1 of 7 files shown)


app/components/operation-area-select-by-name.js
```diff
@@ -0,0 +1,27 @@
+import Component from '@glimmer/component';
+import { inject as service } from '@ember/service';
+import { task } from 'ember-concurrency';
+import { trackedTask } from 'ember-resources/util/ember-concurrency';
+
+export default class OperationAreaSelectByNameComponent extends Component {
```
DamonKennes: The ticket mentions that multiple operation area's should be selectable so I think we need a multiselect for this component. Something similar to OrganizationTypeMultipleSelectComponent

DamonKennes (changes requested):
The filter works for operation area. 

Only have to check to make it multi select as mentioned in the ticket.

DamonKennes (approved):

PR #715, opened 2026-07-13
OP-3817 inhoudelijk thema

DamonKennes:
- Added the new inhoudelijk thema field

TO TEST:
see https://github.com/lblod/app-organization-portal/pull/608


app/components/content-theme-select.js
```diff
@@ -0,0 +1,24 @@
...
+  constructor() {
+    super(...arguments);
+
+    this.loadContentThemesTask.perform();
+  }
+
+  loadContentThemesTask = task(async () => {
+    return await this.store.query('concept', {
+      sort: 'label',
+      'page[size]': 30,
```
Windvis: I think this is missing some entries according to the metadata on the response: `meta	{ count: 32 }`

app/components/content-theme-select.hbs
```diff
@@ -1,17 +1,17 @@
 <div class={{if @error "ember-power-select--error"}}>
   <PowerSelectMultiple
-    @allowClear={{true}}
-    @searchEnabled={{true}}
     @loadingMessage="Aan het laden..."
     @noMatchesMessage="Geen resultaten"
     @searchMessage="Typ om te zoeken"
```
Windvis: Nitpick: This doesn't do anything AFAIK. We _could_ use `@placeholder` if we want to show it when the select is empty but that doesn't match the other fields. I think we should just drop it maybe? Single selects have a `@searchPlaceholder` property, but that isn't used if the search field is part of the trigger (which multiple selects use).

```suggestion
```

app/routes/organizations/organization/core-data/index.js
```diff
@@ -64,6 +65,7 @@ export default class OrganizationsOrganizationCoreDataIndexRoute extends Route {
       scopeLabel,
       kboContact: findPrimaryContact(kboContacts),
       vendors,
+      contentThemes,
```
Windvis, line 68: Is this used somewhere? If not I would just remove it so we don't have to look for usage of it in the future (if we do any refactoring).

You could also use this when displaying the labels though, so you don't need to do the `.hasMany('themes').value()` dance in the getter. Up to you.

app/models/organization.js
```diff
@@ -189,6 +212,17 @@ export default class OrganizationModel extends AgentModel {
...
   }
 
+  get requiresContentThemes() {
+    return !this._hasClassificationId([
+      ...MunicipalityCodeList,
+      ...OCMWCodeList,
+      ...DistrictCodeList,
+      ...ProvinceCodeList,
+      ...WorshipServiceCodeList,
+      ...CentralWorshipServiceCodeList,
```
Windvis: I would maybe extract this array to a variable so it can be reused in both places and we don't accidentally forget to update them both when changes are needed in the future.

Windvis (commented):
Looks good! The only blocker is the page size issue 👍  .

app/routes/organizations/organization/core-data/index.js
DamonKennes: I rewrote the contentThemesString so that it now uses this. This seemed the cleanest imo.

Windvis (approved):

PR #720, opened 2026-07-24
[OP-3844] Add Regionaal zorgplatform and Andere organization types

bdevloed:
cf https://github.com/lblod/app-organization-portal/pull/611

- "Regionaal zorgplatform" mirrors Zorgraad: altLabel "Regionale zorgzone", werkingsgebied + KBO required
- "Andere" gets its own group in the "Type organisatie" filter (new `ORGANIZATION_TYPES.OTHER`) and takes over the legacy generic types' behaviour (contact data editable, site type, membership slots)
- `ASSOCIATION_OTHER` / `CORPORATION_OTHER` removed (codes deleted + orgs migrated by the app PR)
- classification-identifiers test expectations were stale since the OP-3794 types landed — updated to the current taxonomy
- note: `requiresContentThemes` (OP-3817) doesn't exclude the new types, so both require an inhoudelijk thema at creation — shout if "Andere" should be exempt

### DIFF
--- a/app/templates/organizations/new.hbs
+++ b/app/templates/organizations/new.hbs
@@ -219,6 +219,7 @@
               (or
                 this.currentOrganizationModel.isAdministrativeUnit
                 this.currentOrganizationModel.isZorgraad
+                this.currentOrganizationModel.isRegionaalZorgplatform
                 this.currentOrganizationModel.isBosgroep
                 this.currentOrganizationModel.isWoonmaatschappij
                 this.currentOrganizationModel.isPrivateOcmwAssociation
--- a/app/constants/Classification.js
+++ b/app/constants/Classification.js
@@ -52,8 +52,7 @@ export const PevaCodeList = [
   ...PevaProvinceCodeList,
 ];
 
-export const AssociationOtherCodeList = [CLASSIFICATION.ASSOCIATION_OTHER.id];
-export const CorporationOtherCodeList = [CLASSIFICATION.CORPORATION_OTHER.id];
+export const AndereCodeList = [CLASSIFICATION.ANDERE.id];
 
 export const VlaamseGemeenschapscommissieCodeList = [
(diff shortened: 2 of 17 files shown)


app/templates/organizations/new.hbs
```diff
@@ -219,6 +219,7 @@
               (or
                 this.currentOrganizationModel.isAdministrativeUnit
                 this.currentOrganizationModel.isZorgraad
+                this.currentOrganizationModel.isRegionaalZorgplatform
```
DamonKennes, line 222: This adds werkingsgebied when creating this new organization to the create screen. If this is intended, we should add this to the edit/core-detail screens. At the moment werkingsgebied is not shown there.

DamonKennes (changes requested):
Is it intended behaviour that for regionaalZorgplatform and andere the bestuursorganen tab is not shown? If not this should also be changed.

Just asking because via the construct service RAAD_VAN_BESTUUR, ALGEMENE_VERGADERING, and LEIDEND_AMBTENAAR gets added but they are not shown in the frontend.

bdevloed:
> Is it intended behaviour that for regionaalZorgplatform and andere the bestuursorganen tab is not shown? If not this should also be changed.
> 
> Just asking because via the construct service RAAD_VAN_BESTUUR, ALGEMENE_VERGADERING, and LEIDEND_AMBTENAAR gets added but they are not shown in the frontend.

Thx for the review!

It's the existing behaviour for the whole registered-organization family, not something specific to the new types.

The tab is gated by requiresGoverningBodies, which only exists on AdministrativeUnitModel, so Zorgraad/Bosgroep/Woonmaatschappij don't show it either (the governing-bodies route would even crash for them, see the note in routes/.../governing-bodies.js). The construct service has created these organs since OP-3828; showing them for registered organizations is a model + route rewrite covering five types. I'd take that up in a separate ticket.

Werkingsgebied on view/edit: fixed in 9dfad389.

DamonKennes (approved):

PR #721, opened 2026-07-28
[OP-3847] Update to Appuniversum v4

Windvis:
- resolve deprecations
- required dependency bumps

Upgrade guide: https://github.com/appuniversum/ember-appuniversum/pull/566

Probably best to test it as part of the Ember update PR (pending).

PR #722, opened 2026-07-28
[OP-3837] Add a new "Wijkkantoor" site type option for PZ organizations

Windvis:
Builds on #721 since it touches the same code, but only the last commit is relevant for review.

**Test instructions**
- Proxy to your local backend running this PR: https://github.com/lblod/app-organization-portal/pull/612
- Log in as beheerder
- Select a "Politiezone" organization
- Go to "vestigingen" and go to the edit or create view
- Verify that the "Wijkkantoor" option is selectable
- Verify that it's not visible for non-PZ organization types

image

### DIFF
--- a/app/components/address-register-selector.js
+++ b/app/components/address-register-selector.js
@@ -24,16 +24,15 @@ export default class AddressRegisterSelectorComponent extends Component {
     }
   }
 
-  @task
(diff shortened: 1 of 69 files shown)


DamonKennes (approved):
Wijkkantoor works as explained.

PR #723, opened 2026-07-30
[OP-3849] Update to Ember v6.12 LTS

Windvis:
This updates the app to the v6.12 blueprint which is the current LTS. It also does some basic maintenance to get things working as expected and solve most of the linting issues and deprecations. The main remaining one is EmberData/WarpDrive related, but that would be a huge change, so out of scope for this PR.

Functionally, things shouldn't have changed, so a generic test of the app should do it.


DamonKennes (approved):
Did a generic test of the application and I didn't find anything out of the ordinary.

PR #724, opened 2026-08-03
Resolve WarpDrive deprecations and linting errors

Windvis:
During the Ember v6.12 update we silenced some deprecations and warnings because resolving them would be quite the work, and not strictly needed, so we've split it off to a separate PR.


Windvis:
Duplicate of #729 which has a feature build, so closing this one.

PR #725, opened 2026-08-04
[OP-3808] Add a vendor search filter

Windvis:
This lets users find organizations based on the platform they originate from.

**Test instructions**
- Set up a local backend with the following PR: https://github.com/lblod/app-organization-portal/pull/615
- Run the frontend and log in
- Go to the organizations search page and verify that the "Digitaal platform" filter works as expected


app/templates/organizations/index.hbs
```diff
@@ -112,6 +112,14 @@
               @id="statuses"
             />
           </div>
+          <div class="au-o-grid__item au-u-1-2@small au-u-1-1@medium">
+            <AuLabel for="vendor">Digitaal platform</AuLabel>
```
Windvis, line 116: I've used "Digitaal platform" instead of "Bron" (loket) since that's how we already call it in different places in OP. Not sure if that makes sense for the filter though. Bron might still be better?

wolfderechter (approved):

